Issue #11, opened 2016-01-28
github extension should not require nl2br


dholbach:
The use of `nl2br` in the `github` extension is a bit confusing to me as that's not the current behaviour I see on github.com. As an example, check:
* Original .md file: https://raw.githubusercontent.com/ubuntu-core/snapcraft/master/docs/debug.md
* Resulting HTML: https://github.com/ubuntu-core/snapcraft/blob/master/docs/debug.md

facelessuser:
This is because Github annoyingly changed their specs for Github Flavored Markdown without warning: http://stackoverflow.com/a/24576087.  Originally this behavior was not confusing and was actually quite accurate.  I don't know why they would do this as it messes up legacy readme formatting.  Sigh...

Anyways, this puts me in a similar annoying position in order to monkey the Github "feel", I have to do the same thing to others who use this extension.   I need to think about this.  I wonder if it would be better to add a deprecation warning about nl2br before we straight up remove it and mess up rendering of peoples' current documentation without warning.

The github extension is just a convenience extension; there is nothing stopping people from individually including and configuring the same extensions.

I am open to discussion, but I always feel it may anger people to have the github extension without warning rendering their documents in a new format.

dholbach:
Ok, that's fair enough. Thanks for your thoughts on this. For my local use of pymdown I might just go with your suggestion of mixing and matching extensions which "feel" right. :-)

facelessuser:
I think my plan is to add a new setting to not use nl2br which will be off by default,  but have a warning that in the next major release it will be on by default.   After that there will be a warning that the setting will be removed entirely in the future.  This way I can give people a window to get their docs in order.

PR #12, opened 2016-01-28
Remove nl2br from the list of inherited github extensions.

dholbach:
Remove `nl2br` from the list of inherited `github` extensions. It does not reflect reality. As an example compare:

* https://github.com/ubuntu-core/snapcraft/blob/master/docs/debug.md
* https://raw.githubusercontent.com/ubuntu-core/snapcraft/master/docs/debug.md

This closes issue #11.


facelessuser:
As discussed I will be taking a different approach to implement this in 1.1.0.  Thanks for the pull.

PR #13, opened 2016-02-09
Add strip_protocol option to magiclink to hide the http://

bcaller:
so now you can use pymdownx.magiclink(strip_protocol=True) if you want your links to be displayed as [github.com](https://github.com) rather than [https://github.com](https://github.com)

Tests and docs updated.

Not certain if anybody else wants this. Obviously this feature is non-urgent.

### DIFF
--- a/docs/extensions/magiclink.md
+++ b/docs/extensions/magiclink.md
@@ -22,4 +22,9 @@ Just paste links directly in the document like this: https://github.com/faceless
 
 Or even an email address fake.email@email.com.
 
+## Options
+| Option    | Type | Default |Description |
+|-----------|------|---------|------------|
+| hide_protocol | bool | False | If 'True', links are displayed without the initial ftp://, http:// or https:// |
+
 *[GFM]:  Github Flavored Markdown
--- a/pymdownx/magiclink.py
+++ b/pymdownx/magiclink.py
@@ -57,11 +57,14 @@ class MagiclinkPattern(LinkPattern):
         """Handle URL matches."""
 
         el = util.etree.Element("a")
+        el.text = m.group(2)
         if m.group("www"):
             href = "http://%s" % m.group(2)
         else:
             href = m.group(2)
-        el.text = m.group(2)
+            if self.config['hide_protocol']:
+                el.text = el.text[el.text.find("://") + 3:]
+
         el.set("href", self.sanitize_url(self.unescape(href.strip())))
 
         return el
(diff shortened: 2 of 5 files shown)


pymdownx/magiclink.py
```diff
@@ -84,10 +94,24 @@ def handleMatch(self, m):
...
+            ]
+        }
+        super(MagiclinkExtension, self).__init__(*args, **kwargs)
+
     def extendMarkdown(self, md, md_globals):
         """Add support for turning html links and emails to link tags."""
 
-        md.inlinePatterns.add("magic-link", MagiclinkPattern(RE_LINK, md), "<not_strong")
+        md.inlinePatterns.add("magic-link",
+                              MagiclinkPattern(RE_LINK, md, strip_protocol=self.getConfigs()['strip_protocol']),
```
facelessuser: I would prefer if you attached the config to the class as the other extensions do: https://github.com/facelessuser/pymdown-extensions/blob/master/pymdownx/progressbar.py#L251.

tests/extensions/magiclink_strip_http.html
```diff
@@ -0,0 +1,110 @@
+<h3>Should Pass</h3>
```
facelessuser: I really think we don't need to copy the entire test file just to unit test the stripping of the protocol.  You should be able to effectively show it is working with a couple of examples.

docs/extensions/magiclink.md
```diff
@@ -22,4 +22,10 @@ Just paste links directly in the document like this: https://github.com/faceless
 
 Or even an email address fake.email@email.com.
 
+## Options
+| Option    | Type | Default |Description |
+|-----------|------|---------|------------|
+| strip_protocol | bool | False | If 'True', links are displayed without the initial ftp://, http:// or https:// |
+
+
```
facelessuser: Only one empty line is needed here.

pymdownx/magiclink.py
```diff
@@ -84,10 +94,26 @@ def handleMatch(self, m):
...
+                "- Default: False"
+            ]
+        }
+        super(MagiclinkExtension, self).__init__(*args, **kwargs)
+
     def extendMarkdown(self, md, md_globals):
         """Add support for turning html links and emails to link tags."""
 
-        md.inlinePatterns.add("magic-link", MagiclinkPattern(RE_LINK, md), "<not_strong")
+        config = self.getConfigs()
```
facelessuser: What I was really looking for was to attach the config to the MagiclinkPattern object and let it determine when to strip protocol by accessing self.config as was done in the last iteration (just minus all the passing in of the arguments, super, and popping the argument off the param dict).  `el.set("href", self.sanitize_url(self.unescape(href.strip())))` should occur last (even if the stripping won't ever actually affect the sanitizing).

```python
magic_link = MagiclinkPattern(RE_LINK, md)
magic_link.config = self.getConfigs()
md.inlinePatterns.add('magic-link", magic_link, "<not_strong")
```

I don't see it necessary to derive a new class to append 2-3 lines of code, especially when that code is coming after sanitizing when it should come before.  The final URL should be passed in to be sanitized.

pymdownx/magiclink.py
```diff
@@ -57,16 +57,26 @@ def handleMatch(self, m):
...
-        el.text = m.group(2)
         el.set("href", self.sanitize_url(self.unescape(href.strip())))
 
         return el
 
 
+class MagiclinkHideProtocolPattern(MagiclinkPattern):
+    """Convert html, ftp links to clickable links, but without displaying the protocol"""
+
+    def handleMatch(self, m):
```
facelessuser: Need doc string.

pymdownx/magiclink.py
```diff
@@ -57,16 +57,26 @@ def handleMatch(self, m):
...
             href = "http://%s" % m.group(2)
         else:
             href = m.group(2)
-        el.text = m.group(2)
         el.set("href", self.sanitize_url(self.unescape(href.strip())))
 
         return el
 
 
+class MagiclinkHideProtocolPattern(MagiclinkPattern):
```
facelessuser: Need period ending sentence.

bcaller:
Thanks for your patience. I think I finally understood what you meant. You can close this if you don't want it, but it works for Pygments<=2.1.0 now.

facelessuser:
No problem.  It's looking much better now.

PR #17, opened 2016-09-10
Add better support for styling task lists

squidfunk:
The generated source makes it impossible to style checkboxes with custom images or webfonts with pseudo elements, as they are not supported by the CSS specification on input elements.

Placing an empty label next to a checkbox makes this possible, see here:
http://stackoverflow.com/questions/3772273/pure-css-checkbox-image-replacement

An empty label takes no space, as it is an inline element by default. Would be great if this would make it into master, because I would like to incorporate Material Design styled checkboxes into my MkDocs theme: https://github.com/squidfunk/mkdocs-material

### DIFF
--- a/pymdownx/tasklist.py
+++ b/pymdownx/tasklist.py
@@ -34,7 +34,7 @@ RE_CHECKBOX = re.compile(r"^(?P<checkbox> *\[(?P<state>(?:x|X| ){1})\] +)(?P<lin
 def get_checkbox(state):
     """Get checkbox tag."""
 
-    return '<input type="checkbox" disabled%s> ' % (' checked' if state.lower() == 'x' else '')
+    return '<input type="checkbox" disabled%s><label></label> ' % (' checked' if state.lower() == 'x' else '')
 
 
 class TasklistTreeprocessor(Treeprocessor):
--- a/tests/extensions/tasklist.html
+++ b/tests/extensions/tasklist.html
@@ -7,31 +7,31 @@
 <div class="markdown-body">
 <h1>Task List</h1>
 <ul class="task-list">
-<li class="task-list-item"><input type="checkbox" disabled checked> item 1<ul class="task-list">
-<li class="task-list-item"><input type="checkbox" disabled checked> item A</li>
-<li class="task-list-item"><input type="checkbox" disabled> item B<ul class="task-list">
-<li class="task-list-item"><input type="checkbox" disabled checked> item a</li>
-<li class="task-list-item"><input type="checkbox" disabled> item b</li>
-<li class="task-list-item"><input type="checkbox" disabled checked> item c</li>
+<li class="task-list-item"><input type="checkbox" disabled checked><label></label> item 1<ul class="task-list">
+<li class="task-list-item"><input type="checkbox" disabled checked><label></label> item A</li>
+<li class="task-list-item"><input type="checkbox" disabled><label></label> item B<ul class="task-list">
+<li class="task-list-item"><input type="checkbox" disabled checked><label></label> item a</li>
+<li class="task-list-item"><input type="checkbox" disabled><label></label> item b</li>
+<li class="task-list-item"><input type="checkbox" disabled checked><label></label> item c</li>
 </ul>
 </li>
-<li class="task-list-item"><input type="checkbox" disabled checked> item C</li>
+<li class="task-list-item"><input type="checkbox" disabled checked><label></label> item C</li>
 </ul>
 </li>
-<li class="task-list-item"><input type="checkbox" disabled> item 2</li>
-<li class="task-list-item"><input type="checkbox" disabled> item 3</li>
+<li class="task-list-item"><input type="checkbox" disabled><label></label> item 2</li>
+<li class="task-list-item"><input type="checkbox" disabled><label></label> item 3</li>
 </ul>
 <h1>Mixed Lists</h1>
 <ul>
 <li>item 1<ul class="task-list">
-<li class="task-list-item"><input type="checkbox" disabled checked> item A</li>
-<li class="task-list-item"><input type="checkbox" disabled> item B<ol>
+<li class="task-list-item"><input type="checkbox" disabled checked><label></label> item A</li>
+<li class="task-list-item"><input type="checkbox" disabled><label></label> item B<ol>
 <li>item a</li>
 <li>item b</li>
 <li>item c</li>
 </ol>
 </li>
-<li class="task-list-item"><input type="checkbox" disabled checked> item C</li>
+<li class="task-list-item"><input type="checkbox" disabled checked><label></label> item C</li>
 </ul>
 </li>
 <li>item 2</li>
@@ -40,27 +40,27 @@
 <h1>Really Mixed Lists</h1>
 <ul class="task-list">
 <li>item 1<ul class="task-list">
-<li class="task-list-item"><input type="checkbox" disabled checked> item A</li>
+<li class="task-list-item"><input type="checkbox" disabled checked><label></label> item A</li>
 <li>item B
     more text<ol class="task-list">
 <li>item a</li>
 <li>item b</li>
-<li class="task-list-item"><input type="checkbox" disabled> item c</li>
+<li class="task-list-item"><input type="checkbox" disabled><label></label> item c</li>
 </ol>
 </li>
 <li>item C</li>
 </ul>
 </li>
 <li>item 2</li>
-<li class="task-list-item"><input type="checkbox" disabled checked> item 3</li>
+<li class="task-list-item"><input type="checkbox" disabled checked><label></label> item 3</li>
 </ul>
 <h1>Test Tasklists in Subparagraphs</h1>
 <ul class="task-list">
 <li class="task-list-item">
-<p><input type="checkbox" disabled checked> item 1</p>
+<p><input type="checkbox" disabled checked><label></label> item 1</p>
 <ul class="task-list">
 <li class="task-list-item">
-<p><input type="checkbox" disabled checked> item A</p>
+<p><input type="checkbox" disabled checked><label></label> item A</p>
 </li>
 <li>
 <p>item a</p>


facelessuser:
When I put this together, it was just meant to be a simple tasklist like github, though I can see how this might like to be adapted to use custom styling.  It surprises me that something like `label` needs to be added though, and that something like `:before` can't be used (I wish browsers would get their crap together and adopt a common rendering).

Anyways, how common is this `label` approach?  It feels like a bit of a hack...but then again there is probably a lot of web stuff I've seen that could be described like that 😄 .  I know its mentioned in this stackoverflow post, but I am curious how widespread the adoption of this approach is.

squidfunk:
The approach is very common. It's the only way of doing custom checkboxes (or radio buttons). Googling for "custom checkbox" gives the following results:

https://codepen.io/CreativeJuiz/pen/BiHzp
https://css-tricks.com/snippets/css/custom-checkboxes-and-radio-buttons/
http://webdesign.tutsplus.com/articles/quick-tip-easy-css3-checkboxes-and-radio-buttons--webdesign-8953

... all of which use this technique. I do a lot of frontend work and have seen those quite a lot. Why a `label` element and not a let's say `span`? Because normally clicking the label will also toggle the checkbox. Nevertheless: in order to make the checkboxes togglable, each checkbox would need an `id="xyz"` and the label would need a`for= "xyz"`. However, the checkboxes are all disabled, so they aren't clickable anyway, so it's just for styling purposes. If those task list elements SHOULD be clickable in the future (with some background action triggering, regardless of the sense in that), oe would only need to generate ids for the checkboxes and adjust the labels, so it's even future proof.

And yes, browsers suck. It works in Chrome by just defining a pseudo element on the checkbox, but it's not specified so technically it's undefined behaviour.

squidfunk:
Oh, and regardless if you think this is useful or not - I think this project is very useful and well designed and I will promote it on the next version of the Material theme (which is currently in the making and should be released soon). Thanks for the great work!

facelessuser:
> Oh, and regardless if you think this is useful or not - I think this project is very useful and well designed and I will promote it on the next version of the Material theme (which is currently in the making and should be released soon). Thanks for the great work!

Thanks!  

> The approach is very common. It's the only way of doing custom checkboxes (or radio buttons). Googling for "custom checkbox" gives the following results:

This is what I was looking for and it establishes, I think, a fairly good case that this is common.  I'm not really a web developer, so it is always good for me to understand how stuff *actually* gets used outside of my use cases :smile:.

> Nevertheless: in order to make the checkboxes togglable, each checkbox would need an `id="xyz"` and the label would need `afor="xyz"`.

I am not overly concerned about this as Markdown is generally used for generating static HTML like in documentation and such.  Since there is no real syntax to support assigning ids to Markdown lists elements, if this was desired in a document, I would advise doing those elements in straight HTML so that the id and such could manually be defined.  I don't think I would have interest providing modified list syntax to accommodate this.

---

I think I am okay with this, especially since it should be transparent.  I guess It would probably be good if this was mentioned in the documentation.  I don't mind documenting it, but do you mind providing a simple example of CSS using webfonts?  You can post the example here, and I can include it and mention the availability of this in the docs.

squidfunk:
Sounds good to me. I think we should provide an example without icon fonts, because then you have to explain how to include them and it gets complicated quickly. Therefore I suggest we just use the UTF-8 check mark. An average developer should know how to trade them for icon fonts or images. Here is a simple example I threw together quickly:

``` css
.markdown-body .task-list-item {
  position: relative;
}

.markdown-body .task-list-item input[type="checkbox"] {
  opacity: 0;
}

.markdown-body .task-list-item input[type="checkbox"] + label {
  display: block;
  position: absolute;
  top: 50%;
  left: -24px;
  width: 16px;
  margin-top: -8px;
  height: 16px;
  border-radius: 2px;
  background: #CCC;
}

.markdown-body .task-list-item input[type="checkbox"]:checked + label::before {
  display: block;
  position: absolute;
  margin-top: -4px;
  margin-left: 2px;
  font-size: 1.2em;
  line-height: 1;
  border-radius: 2px;
  content: "✔";
  color: #1EBB52;
}
```

bildschirmfoto 2016-09-11 um 10 32 37

facelessuser:
Unicode checkmark is fine.  The more simple the example the better.  I just want to give enough info to demonstrate the concept.

facelessuser:
Hmm. Doesn't seem to nest well...

screenshot 2016-09-11 16 29 23

I will need to play with this to see if this can be resolved

squidfunk:
I will also have a look.

facelessuser:
Okay, I got it:

screenshot 2016-09-11 16 37 29

Had to play with top:

```css
.markdown-body .task-list-item input[type="checkbox"] + label {
  display: block;
  position: absolute;
  top: 12px;
  left: -24px;
  width: 16px;
  margin-top: -8px;
  height: 16px;
  border-radius: 2px;
  background: #CCC;
}
```

facelessuser:
Now that I've played with it and run it through some more rigorous tests, I think I have no other concerns.  Just wanted to make sure it would work when run through more complicated configurations.  I think it works fine.

facelessuser:
Much better when removing the `margin-top` 👍 .  I think I have everything I need.

facelessuser:
I use MkDocs, and the default theme for some reason sets the `label` tag to block.  This breaks styling of the default style task lists.  `label` tags *should* be inline, but I can't guarantee this change will be backwards compatible with everyone's current setup.

I will likely put this under a setting toggle that is disabled by default.  This is just to allow a seamless upgrade for people.  For instance, anyone using MkDocs with the "Read the Docs" theme will likely have broken default tasklist style if I don't.

facelessuser:
Documentation has been added, and the feature is under the setting `custom_checkbox`.

squidfunk:
Okay, that sucks. Then we should do it through an option.

facelessuser:
Interesting conversation over in the linked thread.  While I don't think accessibility is a big deal with tasklists in static documents (probably more important in forms), it might be something to explore and compare the pros and cons.  On release, I'd like to get this right the first time if possible.

squidfunk:
Yes, we could adapt the HTML forms approach, however, I was just striving for minimal changes and the checkboxes are not checkable anyway. What do you think?

facelessuser:
I agree they aren't checkable, so focus isn't a big deal.  Since I have to place it under a feature setting anyways, I open to either.  What is your personal opinion?

squidfunk:
Well, my personal option is that the linked approach is better, so we can adopt it if you like. It also a very, very small step towards actually checkable checkboxes, as we wouldn't need the ids, as stated in the documentation.

However, we cannot wrap the whole content (text) of the task inside the label, because by default, a label is an inline element and the tasklist may contain block elements (p, ul, ...) so the markup could be:

``` html
<li>
  <label class="task-list-state">
    <input type="checkbox" disabled checked />
    <span class="task-list-control"></span>
  </label>
  Feed the cat
</li>
```

Theoretically we could also drop the classes on the `label` and `span`, but it's better to style according to those classes than according to tags.

facelessuser:
> However, we cannot wrap the whole content (text) of the task inside the label, because by default, a label is an inline element and the tasklist may contain block elements (p, ul, ...) and inline elements are not allowed to contain block elements, so the markup should be:

Yes, I had figured that would be the approach.  It is no different than the label approach that is currently in place, and I agree it would be best with this implementation as well.

> Theoretically we could also drop the classes on the label and span, but it's better to style according to those classes than according to tags.

Yes, I prefer keeping the styling as you can never guess how people are going to style their documents.    So it will likely be very similar.  I plan on keeping the styling on `li` and `ol`/`ul` as well.  I rather include classes that don't get used than have to add more later because someone has some use case I didn't think of.  If someone wants to use JS to enable toggling, they will have classes they can target.

I was actually leaning this way a little anyways, so I think I will commit to this updated format.  Since it is under a setting it already it should be a simple change and I will need to update the docs on styling it.

squidfunk:
Okay great, thank you for your cooperation!

Including the class on the wrapping label makes it also possible to account for both cases: a user has enabled the option and another user hasn't - both can be styled independently, which I will do exactly. So that's definitely the way to go.

When do you plan releasing 1.2.0? Are you waiting for more updates? Just so I know when to expect it, no hurry :-)

facelessuser:
I honestly have no planned features in the pipeline right now.  Features are usually based on my current needs, and the project currently fulfills all of my needs.  As long as I don't get any more external requests that I think would be a good fit for the next release, I image there could be a turnaround as soon as this week's end (as long as I don't get too busy with other diversions and work :wink:).

squidfunk:
Okay, if you wish you can defer it a little and I will see if I find some other stuff that could need some customization. However, integrating the other extensions was very frictionless and work's very well so far, so I don't see anything specific coming. I hope I will find the time to finish things quickly and release the reworked version of the theme and we can get both things out :-)

facelessuser:
I'm in no hurry.  As you are dependent on a release here, I will just hold out until you pester me for a release.  The only thing that would force a release earlier is if there was a critical bug that came to light.

facelessuser:
New format is shown below:

```html
<ul class="task-list">
    <li class="task-list-item">
        <label class="task-list-control">
            <input type="checkbox" disabled checked="">
            <span class="task-list-indicator"></span>
        </label>
        item 1
    </li>
</ul>
```

In the docs this is very simple CSS that is used render Unicode checkmarks.

```css
.rst-content .task-list-item {
  list-style-type: none !important;
}

.rst-content .task-list-item input[type="checkbox"] {
  margin: 0 4px 0.25em -20px;
  vertical-align: middle;
}

.rst-content .task-list-control {
  display: inline; /* Ensure label is inline incase theme sets it to block.*/
}

.rst-content .custom-task-list .task-list-control {
  position: relative;
  display: inline-block;
  color: #555;
  cursor: pointer;
}

.rst-content .custom-task-list .task-list-control input[type="checkbox"] {
  position: absolute;
  opacity: 0;
  z-index: -1; /* Put the input behind the label so it doesn't overlay text */
}

.rst-content .custom-task-list .task-list-indicator {
  position: absolute;
  top: -8px;
  left: -18px;
  display: block;
  width:  14px;
  height: 14px;
  color: #eee;
  background-color: #eee;
  border-radius: .25rem;
}

.rst-content .custom-task-list .task-list-control input[type="checkbox"]:checked + .task-list-indicator::before {
  display: block;
  margin-top: -4px;
  margin-left: 2px;
  font-size: 16px;
  line-height: 1;
  border-radius: 2px;
  content: "✔";
  color: #1EBB52;
}
```

squidfunk:
LGTM! I will keep you updated.

squidfunk:
I'm approaching the release of version 1.0.0 for my theme now, for which it would be great to have the possibility to style the task lists. Could you push the feature out in a release? PyMdown will be mentioned on the docs for the theme.

facelessuser:
I will hopefully get to this by the end of the week.

facelessuser:
Released

squidfunk:
Thanks! I will keep you updated, when the RC is out.

PR #58, opened 2017-02-13
Snippets - changed "appending" to "prepending"

seanmadsen:


facelessuser:
Good catch.

facelessuser:
Could you add prepending to the dictionary so the spelling test passes?

seanmadsen:
Now passing. Very cool that you have a spelling test! :ok_hand:

facelessuser:
It is likely that prepend and some of the other forms are already in aspell's main dictionary.  I usually only add words that cause failure, but this is fine.  Thanks for the pull.  It doesn't matter how many times I read over the docs, I always miss something :).

PR #149, opened 2017-11-14
Feature/mention repo

facelessuser:
Adds the ability to mention a repo `@user/repo`, but will output the link visually as `user/repo` or just `repo` if the user is the default configured user.

Also make title labels configurable.

And add provider classes.


facelessuser:
I have a feeling Travis failing on merge commits occasionally is somehow related to new lines.  If this is the case, I may need to ensure in the future that HTML files are always generated with Unix newlines, because this is annoying.

PR #176, opened 2017-12-02
Rename plainhtml and update strip_attributes

facelessuser:
Rename plainhtml to striphtml and change strip_attributes to be a list that is empty by default.


facelessuser:
FYI `pymdownx.plainhtml` will still be available for somewhere around 6 months - 1 year.  It will also take it's options in the legacy format. All of this is for backwards compatibility. In the 5.0 release, `pymdownx.plainhtml` will be completely removed in favor of `pymdownx.striphtml`.

Issue #187, opened 2017-12-08
Math content padding incorrectly with latest Material theme


facelessuser:
Need to regen with Material 2.2.0 as current documentation is messed up.  Last known good version is 2.2.0. This is pre-admonition/details fixes.

Extra padding with nested details/admonitions. Though 2.2.2 seems to fix this, it also adds `overflow: auto` which can make scroll bars appear in weird cases.

<img width="664" alt="details" src="https://user-images.githubusercontent.com/1055125/33785646-385da55e-dc23-11e7-841e-834d0f0afc8a.png">

Math getting cut off.  This gets cut off in both 2.2.1 and 2.2.2.

<img width="668" alt="arithmatex" src="https://user-images.githubusercontent.com/1055125/33785656-3fa6775a-dc23-11e7-8a57-179543f450af.png">

@squidfunk, just an FYI.  I'll create an issue over on Material once I have time to look into this more.

facelessuser:
Seems the padding on the last child is causing the issue, I just don't understand why yet.

```css
html .md-typeset .admonition :last-child, html .md-typeset details :last-child {
    margin-bottom: 0;
    padding-bottom: 1.2rem;
}
```

facelessuser:
Currently Material is assigning any element that is a *last-child* with padding.  This is bad.  So the math elements have some children who or *last-child* to their parent even if they are not direct children. of `.admonition`.  Since they are descendants, they inherit the padding.

Something like `.admonition>:last-child` and `.details>:last-child` would be more appropriate. as it would limit the padding the last *direct* child.

facelessuser:
This is definitely an upstream issue, so we will just gen with 2.2.0 until a future release rectifies this.

facelessuser:
Documents have been regenerated with 2.2.0.  I will leave this open as a reminder to remove the lock on the version once Material addresses this CSS issue.

facelessuser:
Issue is fixed.  We can now remove the version lock on Material.  Once that is removed, we can close this.

PR #200, opened 2017-12-19
 Use external pyspelling tool

facelessuser:
Now have an external module to replace spellcheck.py.  Use pyspelling to automate spellchecking.


facelessuser:
This also removes material limit of later versions and requires 2.2.5+ which resolves #187.

PR #208, opened 2017-12-28
[Add][Progressbar] 10, not 20 increment

Kristinita:
### 1. Note

I understand, that I need to edit another files for pull-request. But I think, that you can decline this pull-request and my work will be wasted. If you accept this pull-request, I will edit another files.

### 2. Behavior before pull request

Before

### 3. Behavior after pull request

After

### 4. Argumentation

I think, that 20 increment — it a little. Personally, I need more detailed progress bar and more optional colors.

### 5. Testing environment

+ Python 3.6.3,
+ Pelican 3.7.1,
+ Python-Markdown 2.6.10,
+ pymdown-extensions 4.7.

### 6. Alternatives

If you don't accept this pull-request, it would be nice, if you add 10 increment as an option.

Thanks.

### DIFF
--- a/pymdownx/progressbar.py
+++ b/pymdownx/progressbar.py
@@ -4,7 +4,6 @@ Progress Bar.
 pymdownx.progressbar
 Simple plugin to add support for progress bars
 
-~~~
 /* No label */
 [==30%]
 
@@ -13,7 +12,6 @@ Simple plugin to add support for progress bars
 
 /* works with attr_list inline style */
 [==50/200  MyLabel]{: .additional-class }
-~~~
 
 New line is not required before the progress bar but suggested unless in a table.
 Can take percentages and divisions.
@@ -21,9 +19,8 @@ Floats are okay.  Numbers must be positive.  This is an experimental extension.
 Functionality is subject to change.
 
 Minimum Recommended Styling
-(but you could add gloss, candy striping, animation, or anything else):
+(but you could add gloss, candystriping, animation or anything else):
 
-~~~.css
 .progress {
   display: block;
   width: 300px;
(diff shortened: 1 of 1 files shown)


pymdownx/progressbar.py
```diff
@@ -4,7 +4,6 @@
 pymdownx.progressbar
 Simple plugin to add support for progress bars
 
-~~~
```
facelessuser: Please put the `~~~` back as they are used as fences to prevent the spell checker from checking this region of the docstring.

facelessuser:
First, thanks for the pull! In general, I'm okay allowing the progress increment to be user configurable, but we need to do it in a way that doesn't break users who are already using this.  Also, we need to make sure the tests are not failing.  Please see my review comments and address accordingly.  Once the tests are cleared up, and this is implemented in a way that is configurable, but by default works the same as the old way, we can merge this.

pymdownx/progressbar.py
```diff
@@ -107,11 +106,12 @@
 DEALINGS IN THE SOFTWARE.
 """
 from __future__ import unicode_literals
+from . import util
 from markdown import Extension
-from markdown.inlinepatterns import Pattern, dequote
 from markdown import util as md_util
 from markdown.extensions.attr_list import AttrListTreeprocessor
-from . import util
+from markdown.inlinepatterns import Pattern
```
facelessuser: I know some linters may complain about importing multiple objects on the same line, but I personally prefer these to be imported on the same line.

facelessuser:
I've created a new pull which addresses this properly: #210.  I will close this issue in favor of the new #210 pull.  Thanks for the suggestion!

PR #249, opened 2018-02-27
Add option to let the user check/uncheck the checkboxes

g0rk4:


facelessuser:
I'm not sure I follow what this pull is trying to solve, can you explain?

facelessuser:
1. Okay, let me just give some feedback. In general, this is used to create static pages, but I get the feeling you want to do something more dynamic.  Why not use JavaScript?  Can you elaborate why you think this is needed?

2. Tests are currently failing, so that would need to be fixed assuming I decide to accept this.

3. Documentation would need to be updated to describe the new feature (again assuming I decide to accept this).

g0rk4:
Hi @facelessuser,

First of all thank you for your quick reply !

Indeed I need this functionnality in order to be able to check dynamicaly some checkboxes. For exemple, I write some documentations of installation tools and I want to let the final user to follow the procedure by checking the steps already done. 

If you will accept this pull request, I'll fix all the things you mentioned :

facelessuser:
I guess I'm okay with this, but all the mentioned things will have to be addressed first.  New feature should be tested, all tests must pass, and there needs to be documentation.

g0rk4:
Hi @facelessuser , i made changes on PR : 

- Add option to let the user check/uncheck the checkboxes;
- Test cases added;
- Update documentation;

And now i'll try to understand why the tests are falling on your CI

docs/src/markdown/extensions/tasklist.md
```diff
@@ -85,7 +99,7 @@ Classes               | Description
     }
     ```
 
-??? settings "Custom Tasklist CSS"
```
facelessuser: This is a details object, not a header.  Please change the syntax back to how it was.

docs/src/markdown/extensions/tasklist.md
```diff
@@ -63,14 +63,28 @@ If `custom_checkbox` is enabled the structure will be as follows:
...
+```
+
+| Classes               | Description                                                                                            |
+| --------------------- | ------------------------------------------------------------------------------------------------------ |
+| `task-list`           | Attached to either the `ul` or `ol` tag and represents the entire list element.                        |
+| `task-list-item`      | This is attached the `li` tag and represents an item in the list.                                      |
+| `task-list-control`   | This is attached to the `label` tag and represents the control object.                                 |
+| `task-list-indicator` | This is attached to the `span` directly following the input and is used to style the visual indicator. |
+
+### settings "Basic Tasklist CSS"
```
facelessuser: This should be a details object, not a header.  Please change the syntax to how it was.

docs/src/markdown/extensions/tasklist.md
```diff
@@ -141,9 +155,10 @@ Classes               | Description
 
 ## Options
 
-Option            | Type | Default      | Description
------------------ | ---- | ------------ | ------------
-`custom_checkbox` | bool | `#!py3 False` | Generate task lists in such a way as to allow for styling the check box with CSS.
+| Option               | Type | Default       | Description                                                                       |
```
facelessuser: Please don't rewrite table syntax that I am using.  Please remove leading and trailing pipes (`|`).  I understand some may prefer to have tables flanked by those, but I do not.

pymdownx/tasklist.py
```diff
@@ -30,17 +30,18 @@
...
-
-def get_checkbox(state, custom_checkbox=False):
+def get_checkbox(state, custom_checkbox=False, clickable_checkbox=False):
     """Get checkbox tag."""
 
     if custom_checkbox:
         return (
             '<label class="task-list-control">' +
-            '<input type="checkbox" disabled%s/>' % (' checked' if state.lower() == 'x' else '') +
+            '<input type="checkbox"%s%s/>' % (' disabled' if clickable_checkbox == False else '',
```
facelessuser: Tests are failing because you added a trailing self closing `/` in the input tag.  This is probably better so that when people insert this into xhtml (but who is doing this these days).  So I'm fine with the change, but you need to regenerate changed, failing tasklist tests.

facelessuser:
I've noted in the review why CI is failing.  Also, I've noted additional changes that need to fixed.

g0rk4:
Sounds good, thank you @facelessuser. 

Tell me if something else is needed

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
facelessuser: Please remove .vscode/settings.json file.  Feel free to exclude it from the repo via .gitignore.

facelessuser:
There was actually no need to remove the `/` in tags.  As I mentioned it is better for xhtml.  I was more mentioning we needed to regen existing tests not new tests.  But maybe we can just do that as a separate issue.

g0rk4:
Hi @facelessuser,

I don't understand about the `/` because I do not removed any one in tags.

.vscode/settings.json
g0rk4: Done, i'll push changes when we wil be ok with the `/`

facelessuser:
I was referring to the fact that you changed ot `'<input type="checkbox"%s%s/> ', and then back to `'<input type="checkbox"%s%s> '`. `/>` is more appropriate as this could be used in xhtml. I cannot look at the history because you force pushed, which rewrote the history, but that was originally why it was failing.

For custom tasklists it is still setting `/>`, but for normal tasklists, it is not. It should be the same for both.  I would prefer it to be the same for both. That is what I was saying.  Technically, this is a bug on master, but you fixed it, and then reverted it.  I was mentioning earlier that it was good that you fixed it, and you just needed to regenerate the failing tests to fix the CI failures.

g0rk4:
All right, now I get it !

I'll fix it this evenning

g0rk4:
Ok, it's done

facelessuser:
Thanks, I'll try this out some time soon, but so far, everything looks good.

g0rk4:
My pleasure

facelessuser:
Went ahead and downloaded it.  Everything looks in order.  This will be in `4.9.0`.

Issue #215, opened 2018-01-08
Prepare for Python Markdown inline processor changes


facelessuser:
I am currently reworking inline processor in a new InlineProcessor class for Python Markdown which improves inline performance, but it also adds the ability to do more analysis and alter the bounds of the original match (ref https://github.com/Python-Markdown/markdown/issues/551).  This is much more flexible. For example, we can match just `[` for links, and then in the plugin tokenize the next characters handling recursive `[]` which Python's re cannot handle.  Python Markdown is moving all inline processors in that direction and providing the legacy versions of the old methods for the basic patterns, but there are two `InlineProcessors` we use that are not going to be accessible after the change `LinkPattern` and `HtmlPattern`.  So we need to simply prepare by checking now whether we can import `LinkPattern` in MagicLinks, and `HtmlPattern` in SmartSymbols.

We will provide alternate logic for these two cases (import succeeds and import fails).  All other patterns can use legacy logic for now. After the Python Markdown finally releases the change, we can change all of pymdown-extensions to utilize the new `InlineProcessor` classes and drop the `Pattern` classes.  Then we can lock pymdown-extensions to whatever version of Markdown that introduces the new `InlineProcessor` classes.

So basically there are two phases:

- [x] Compatibility phase -- do the bare minimum to keep working (make sure we work now, and once the changes are released)
- [ ] Full implementation phase (we fully use the new InlineProcessor class and abandon Pattern).

facelessuser:
This will be handled under #351.

PR #266, opened 2018-03-26
Increase required Markdown version to 2.6.10

lwfitzgerald:
The magiclink extension uses ancestry exclusion to prevent auto-linking
when text is already inside a link. Support for this was only added
to Markdown in version 2.6.10 (see Python-Markdown/markdown#598).

### DIFF
--- a/requirements/project.txt
+++ b/requirements/project.txt
@@ -1,1 +1,1 @@
-Markdown>=2.6.0
+Markdown>=2.6.10


facelessuser:
Yes, I am aware of the ancestry addition as I added it on Python Markdown and here in pymdown-extensions.  But it is a backwards compatible change, meaning pymdown-extensions will work with or without it (though you will not get the benefit of the ancestry check without the latest Python Markdown).

I don't *usually* force specific versions unless there is good reason to force it.  People sometimes have reasons they don't want to update a version. Whenever the next Python Markdown releases, I plan on forcing pymdown-extensions to use the latest Python Markdown as I will be reworking inline patterns to take advantage of the new inline methodology #215.  There will be no fallback for backwards compatibility, so a force will be necessary.

lwfitzgerald:
Although the change is backwards compatible, the test suite currently asserts the output matches that of an run with `Markdown>=2.6.10` installed.

When I ran the tests in an environment with `Markdown==2.6.8`, the tests failed. This doesn't seem correct as the environment met the package requirements.

As the backwards compatibility is intentional, I think the test suite needs a small tweak to test against output dependent on the version of Markdown installed.

facelessuser:
Ahh, that is a fair point.  In CI, the latest is always pulled, so this hasn't been an issue, but I can see how it would be an issue in an environment that isn't clean and installing the first the time. Either I would need  a separate testing requirements file, or update the project requirements...

We'll go ahead and pull this then.

Issue #301, opened 2018-06-01
SuperFences: Tabbed Fenced Code


facelessuser:
See https://github.com/squidfunk/mkdocs-material/issues/778.

facelessuser:
Ongoing work is found in this pull: https://github.com/facelessuser/pymdown-extensions/pull/302.

PR #302, opened 2018-06-01
Initial fenced tabs work (WIP)

facelessuser:
Ref #301


facelessuser:
It has recently become aware to me that maybe our generated ids should be more obscure. Ref https://github.com/squidfunk/mkdocs-material/issues/790.

PR #336, opened 2018-07-13
Minify Browser Support Code

Richienb:
Minified the browser support code using [CSSO](https://css.github.io/csso/csso.html) and [Closure Compiler](https://closure-compiler.appspot.com/home) (both recommended by Google) but still provided the simplified and beatified version. This is useful for a number of reasons:

- Code is much faster to load
- Code is much more lightweight
- Google pagespeed gives it a higher score and boosts SEO etc.


facelessuser:
Hi @Richienb, you make some very good points about page load, and minimizing is indeed a very good practice. Our official documentation even minimizes our [styles and scripts](https://github.com/facelessuser/pymdown-extensions/tree/master/docs/theme).

In general, the documentation isn't really meant to provide final, production quality styles and scripts, just simply to give people something that technically works and is a good starting point, so readable code with comments is much preferable. It is expected that the user may tweak and minimize for their own personal use.

It's pretty trivial for people to quickly minimize the examples as there are plenty of online minimizers for [CSS and JavaScript](https://www.minifier.org/), so providing minimized versions isn't really something I am currently aiming to provide.

I appreciate your interest and your contribution, and though I will probably pass on this particular pull, I look forward to any future contributions you may have that could be of benefit. Thanks for your interest.

Richienb:
Ok. Thanks for that feedback!

Issue #349, opened 2018-08-01
[Superfences] Pass fenced code block config (hl_lines, linenums, ...) to custom formatter func


dbader:
Hi @facelessuser, I absolutely love your MD extensions. They're powering large parts of realpython.com and your documentation has been a joy to work with 😃 

I'm using the custom formatter functions in `SuperFences` to customize rendering and syntax highlighting. I'm looking to find a way to access fenced code block config like `hl_lines` and `linenums` from the formatter function.

As best as I can tell, a small change in the custom formatter function API to start passing through those config values would achieve this quite easily:

https://github.com/facelessuser/pymdown-extensions/blob/881e852ecfd06449fe608932c3abd4deffe2005e/pymdownx/superfences.py#L370

```python
    def process_nested_block(self, ws, content, start, end):
        """Process the contents of the nested block."""

        self.last = ws + self.normalize_ws(content)
        code = None
        for entry in reversed(self.extension.superfences):
            if entry["test"](self.lang):
                #  
                # TODO: Pass extra data like self.hl_lines to entry['formatter']:
                # 
                code = entry["formatter"](self.rebuild_block(self.code), self.lang)
                break
```

Do you think this is the best way to go about it or should I look elsewhere? Happy to put together a PR if you think the approach I suggested makes sense.

Thanks again!

— Dan

facelessuser:
Hmm. I guess I always looked at the custom formatter as something for things like UML or other non-code like things.  In those cases, `hl_lines` and `linenums` have no meaning.  And since you can't really define custom options, it didn't make since to pass such things...

I'm going to make a guess, is this to provide a custom format for a JavaScript highlighter?

dbader:
> I'm going to make a guess, is this to provide a custom format for a JavaScript highlighter?

Close! I'm doing some transforms on the source code inside the fenced block, like removing / adding `>>>` interpreter prompts and then pass it to `pymdownx.highlight.Highlight.highlight()` to render the syntax highlighted HTML.

As I'm writing this I realized I could put the config data *into* the code block and then parse it out in the custom formatter:

    ```custom_lang
    #$ hl_lines="1 2 3"
    lorem ipsum dolor
    sit amet ...
    ```

Thanks for the rubber ducking session lol... It would still be nice to standardize the syntax and have access to the config info in the custom formatter func, but I'd be equally happy with a workaround.

facelessuser:
If I was approaching this, there are a couple of things I would take into consideration.

[...]

PR #350, opened 2018-08-02
Allow custom options (WIP)

facelessuser:
Allow custom options and optionally pass them to the formatter.

Ref #349

### DIFF
--- a/pymdownx/superfences.py
+++ b/pymdownx/superfences.py
@@ -35,6 +35,7 @@ from markdown.preprocessors import Preprocessor
 from markdown.postprocessors import Postprocessor
 from markdown.blockprocessors import CodeBlockProcessor
 from markdown import util as md_util
+import functools
 from . import highlight as hl
 from . import util
 import re
@@ -158,6 +159,24 @@ def default_validator(language, options):
     return len(options) == 0
 
 
+def _validator(language, options, validator=None):
+    """Validator wrapper."""
+
+    return validator(language, options)
+
+
+def _formatter(source, language, options, pass_opts=False, class_name="", fmt=None):
+    """Formatter wrapper."""
+
+    return fmt(source, language, class_name, options) if pass_opts else fmt(source, language, class_name)
+
+
+def _test(language, test_language=None):
+    """Test language."""
+
+    return test_language is None or language == test_language
+
+
 class SuperFencesCodeExtension(Extension):
     """SuperFences code block extension."""
 
@@ -191,7 +210,7 @@ class SuperFencesCodeExtension(Extension):
         self.superfences.append(
             {
                 "name": name,
-                "test": lambda l, language=name: language == l,
+                "test": functools.partial(_test, test_language=name),
                 "formatter": formatter,
                 "validator": validator
             }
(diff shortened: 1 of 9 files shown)


facelessuser:
## Todo

- [ ] Need unittests
- [ ] Possibly replace lambda with functools.partial
- [ ] Documentation

pymdownx/superfences.py
```diff
@@ -46,22 +46,18 @@
...
-        (?:[ \t]+(?P<linestep>[\d]+))?                                        #   Line step
-        (?:[ \t]+(?P<linespecial>[\d]+))?                                     #   Line special
-    (?P=quot2)[ \t]*|
-    (?P<tab>tab=)(?:(?P<quot3>"|')(?P<tab_title>.*?)(?P=quot3))?[ \t]*        # Tab specifier
-    )*
-    }?[ \t]*$                                                                 # Language closing
+    (?P<fence>~{3,}|`{3,})[ \t]*                                                          # Fence opening
+    (\{?                                                                                  # Language opening
+    \.?(?P<lang>[\w#.+-]*))?[ \t]*                                                        # Language
+    (?P<options>(?:\b[a-zA-Z][a-zA-Z0-9_]*=(?:(?P<quot>"|')[^"'\n\r]*(?P=quot))?[ \t]*)*) # Options
```
facelessuser: This is evaluated line by line, so it is probably not necessary to exclude `\n\r`....not sure what I was thinking here.

dbader:
From an enduser perspective I really like this. Implementing a custom validator seems straightforward and looking at your updated `highlight()` implementation acting on the custom config later on works the way I'd expect. I'm a 👍 This is cool!

dbader:
This might be helpful: https://stackoverflow.com/questions/4764547/creating-dictionary-from-space-separated-key-value-string-in-python, especially the `shlex` example. I think it should be backwards compatible with the current regex-based `hl_lines=...` parser.

pymdownx/superfences.py
```diff
@@ -134,6 +130,34 @@ def fence_div_format(source, language, css_class):
...
+    # Check format of valid keys
+    if okay:
+        for opt, validator in (('hl_lines', RE_HL_LINES), ('linenums', RE_LINENUMS)):
+            if opt in options:
+                value = options[opt]
+                if value is None or validator.match(options[opt]) is None:
+                    okay = False
+                    break
+
+    return okay
```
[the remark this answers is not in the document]
facelessuser: Yeah, its a habit of mine to generally have one exit point.  For better or for worse :smile:.  There are times I deviate from this, but it is usually something I do. If it was a long chain of multiple `if okay`, I'd probably just start returning as you suggested.

pymdownx/superfences.py
```diff
@@ -556,19 +606,33 @@ def search_nested(self, lines):
                 lines = lines[:start] + [fenced] + lines[end:]
         return lines
 
-    def highlight(self, src, language):
+    def highlight(self, src, language, options=None):
```
facelessuser: It's weird to have options passed in but use `self.options`, I should just use the passed in one.

facelessuser:
`shlex` is indeed a good suggestion, but I generally don't like to pull in libraries when the solution is already quite trivial.  If it cleans up an overly complicated solution, or provides something extra that I feel I need but am missing, than I'm usually on board, but in this particular case I think it may be a little overkill.

facelessuser:
Glad to hear! Hopefully I'll get documentation updated soon and get a released out.

PR #360, opened 2018-08-15
Linenums Enhancements

facelessuser:


facelessuser:
Main issue here is we need tests and documentation.

PR #387, opened 2018-09-20
If text is returned, html chars will be escaped

facelessuser:


facelessuser:
Thinking about this a bit more, some thoughts have come to me about this.  Content is inside a code block, so should we allow text to be returned out of this that other extensions and steps can process?  But then if we don't, when a user returns an ElementTree object, this will allow other extensions to operate on it unless the content is wrapped in AtomicString. Should I care? Or just make a note in documentation?

A separate thing brings up the fact that none of the formatters in this and SuperFences has no access to the `md` object.  Should we pass in `md`? I'd have to think about how to approach this in a non-breaking way, or at least a non-breaking way in the short term.

facelessuser:
We are going to keep the convention that what is returned is expected to be HTML ready.  The user is responsible for escaping what they think is important to escape in their return string.

As far as ElementTree return, there is no way for me to be sure, what should be wrapped in Atomic strings, what should be HTML stashed, and what the user didn't intend to protect.

We will document that strings will be stashed, and the user **should** wrap HTML content in either Atomic strings, and stash any HTML strings as the ElementTree element is not intended to be processed further, but people can do what they want 🤷‍♂️ .

Issue #396, opened 2018-09-28
Roadmap 6.0


facelessuser:
There will be breaking changes, but hopefully only to SuperFences custom format.

- [x] Superfences will have an updated format that always takes options and the md object. Passing in the md object (new), will give access to markdown related functions such as `md.htmlStash` which can be useful, or even `md.meta` if your solution requires meta data.  It's worth just breaking things for a 6.0 release so we don't have more "special case" code like we do with "options". To implement this in a backwards compatible way, we'd need to require some kind of additional option in a custom fence to say pass me the markdown object.

- [x] Since we are breaking things 😄, SuperFences will no longer include `flow` and `sequence` by default.  This is a niche setup. Not everyone using SuperFences is using the UML feature, and not everyone using UML is using it like the default is.  Maybe they are using mermaid.  So by default, custom fences will be empty.

- [x] InlineHilite will allow custom inline spans.  The work for this is already mostly done, but I have some tweaks.  I want to add access to the md object just like SuperFences has.

- [x] Fully compliant with the new methodology of Python Markdown 3.  Take advantage of the new stuff, and move away from the deprecated.  Require Python Markdown 3 by default.

facelessuser:
I'm going to nix the inline options idea. After thinking about how to implement it, I realize now I don't want to muddy the inline syntax more than it is.  There is no reason a user couldn't just implement there own option header in their own custom inline.  There is no need for me to do and possibly break other people's expectations.  I don't mind breaking an option when it makes sense, but I'm not really keen on breaking the syntax. I still want to feed in the md object though.

facelessuser:
In the end, I think the real broken part is going to be people with existing custom fences.  The format will change, but it is generally for the better.  The only other people are those that relied on the default UML option.  They will have to manually define it moving forward.

facelessuser:
We shuffled and reworked some priority of extensions to address behavior with other extensions (problems that have existed for a while, but probably not experienced excepted in narrow cases).  So generally planned stuff is done.  I will be addressing some other bug related things, but I am not planning in other features in the 6.0.0 release.

Issue #351, opened 2018-08-03
Python Markdown 3 support


facelessuser:
Work towards eliminating all deprecations introduced in Python Markdown 3 and take advantage of new methodologies.

facelessuser:
This will address issue #215 as well.

PR #398, opened 2018-09-29
Python Markdown 3 Updates

facelessuser:
Ref #396 #351

### DIFF
--- a/docs/src/markdown/changelog.md
+++ b/docs/src/markdown/changelog.md
@@ -1,9 +1,13 @@
 # Changelog
 
-## 5.1.0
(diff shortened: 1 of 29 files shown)


facelessuser:
If we can, we should postpone SuperFences' preserved tab conversion until after normalize whitespace like the non preserved tab does.  As we've added support for passing in the md object into the formatters so users can take advantage of meta data etc., preserved tab occur before such data is captured.  If we can just preserve the source with raw white space, and then restore it/handle it at the same time other fences are processed, it should allow things to work roughly the same.

PR #523, opened 2019-03-05
Fix extra extensions list, add note about github extension deprecation

shvchk:


docs/src/markdown/usage_notes.md
```diff
@@ -13,16 +13,17 @@ Also, you shouldn't include an extension more than once. If you try to include m
...
+    'markdown.extensions.footnotes',
+    'markdown.extensions.attr_list',
+    'markdown.extensions.def_list',
+    'markdown.extensions.tables',
+    'markdown.extensions.abbr',
+    'pymdownx.extrarawhtml'
     ```
 
-- `pymdownx.github` doesn't replace any extensions and is just a convenience extensions to include most of the extensions needed to get a GitHub-ish feel.  Remember to avoid including this and then including conflicting extensions or doubles of extensions.  Here is the full list of included extensions:
+- `pymdownx.github` (deprecated) doesn't replace any extensions and is just a convenience extensions to include most of the extensions needed to get a GitHub-ish feel.  Remember to avoid including this and then including conflicting extensions or doubles of extensions.  Here is the full list of included extensions:
```
facelessuser: This entire statement should be removed. GitHub extension has been entirely removed. It isn't deprecated, it is gone.

PR #527, opened 2019-03-08
Add figure and figcaption handling

funk1d:
Add `<figure>` and `<figcaption>` handling.

- `%%%`  start a `<figure>`
- `%:`  start a `<figcaption>` in `<figure>`

If `` is the only thing in `<p>`, the `<p>` tag will be got rid of.

Example:
```
%%%

%%% figure-class here
    ![img-alt](/source/of/img.jpg){: img-attributes here}

    %: figcaption here
    {: figcaption-attributes here}
```
will be rendered as:

```html
<p>%%%</p>
<figure class="figure-class here">
<img alt="img-alt" here="here" img-attributes="img-attributes" src="/source/of/img.jpg" />
<figcaption figcaption-attributes="figcaption-attributes" here="here">figcaption here</figcaption>
</figure>
```

Starter only will not be rendered as `<figure>`.
The attributes render is supported by the origin [Python-Markdown](https://python-markdown.github.io/)'s  [`attr_list`](https://python-markdown.github.io/extensions/attr_list/) extension.

### DIFF
--- a/pymdownx/fig_cap.py
+++ b/pymdownx/fig_cap.py
@@ -44,7 +44,8 @@ class FigureBlockProcessor(BlockProcessor):
         """Test block."""
 
         sibling = self.lastChild(parent)
-        return (self.START.search(block) or (block.startswith(' ' * self.tab_length) and sibling is not None and sibling.tag == 'figure'))
+        return (self.START.search(block) or 
+                (block.startswith(' ' * self.tab_length) and sibling is not None and sibling.tag == 'figure'))
 
     def run(self, parent, blocks):
         """Convert to figure block."""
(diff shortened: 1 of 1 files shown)


facelessuser:
Pymdown Extensions is a collection of extensions that I mainly wrote for myself. While I generally am open to suggestions for existing extensions, and on rare occasions open to possibilities of new extensions, I generally don't accept additional extensions as it creates more work for me as I must then support them.

I mainly support extensions that I am regularly using, and while I appreciate your interest to contribute to Pymdown Extensions project, I'm afraid I'm not looking to support this kind of extension. I think it might be a great extension that you could support in your own repo though. I imagine there may be other like minded people who'd like such an extension.

funk1d:
Thanks for your replying. I may make it a third party extension then. Very appreciate your work on this project!

PR #687, opened 2019-10-12
Initial work for custom enterprise repositories

### LINKED ISSUE #583: MagicLink GitHub Enterprise Support
# Request

I'd like to be able to reference GitHub Enterprise repositories in a similar fashion to how one would reference GitHub repositories. The same might as well work for GitLab and BitBucket.

## Usage

I want to be able to define custom providers, e.g, `github`, `pymdownx`, and `mirror` which could be referenced via `@github`, `@pymdownx`, or `@mirror`.

So if I did something like `@mirror` I would expect a link something like this to be generated `[mirror/pymdown-extensions](https://github.enterprise.com/mirror/pymdown-extensions)`.

And if I set the default provider to `pymdownx` and did `#1` I would get `[facelessuser/pymdown-extensions](https://github.com/facelessuser/pymdown-extensions/issues/1)`.

# Proposal

1. Allow users to pass in options to define custom providers.
2. Redefine the current model of a `provider` as more of a template which reflects how data should be parsed and represented.

## Identified Configuration Changes

I think practically what that would mean is changing the configuration options to look something like this:

```yaml
# This is the default provider to use.
provider: pymdownx

# This is a collection of providers you provide.
providers:
  # This is the github provider, I'm listing it explicitly as an example
  # but it should exist by default. The same is true for gitlab and bitbucket.
  # Using an identical name would merge properties. 
  - name: GitHub
    provider: github
    url: https://github.com
  
  # This is this repository!
  - name: Python Markdown Extensions
    provider: pymdownx
    template: github
    user: facelessuser
    repo: pymdown-extensions 

  # This is a mirror of pymdownx hosted by enterprise.com.
  - name: A Python Markdown Extensions Mirror
    template: pymdownx
    url: https://github.enterprise.com/[maybe_some_path?]
    user: mirror
```

## Identified Code Changes

There's a few code changes I've immediately identified and considered in order to support this request.
I think most of these changes will revolve around the way the regular expressions are defined currently.

### `PROVIDER_INFO`

The current `PROVIDER_INFO` info should be composed of just shape information.

```py
PROVIDER_INFO = {
    "gitlab": {
        "provider": "GitLab",
        "url": "https://gitlab.com",
        "user_pattern": RE_GITLAB_USER,
        "issue": "%s/%s/issues/%s",
        "pull": "%s/%s/merge_requests/%s",
        "commit": "%s/%s/commit/%s",
        "compare": "%s/%s/compare/%s...%s",
        "hash_size": 8
    },
    "bitbucket": {
        "provider": "Bitbucket",
        "url": "https://bitbucket.org",
        "user_pattern": RE_BITBUCKET_USER,
        "issue": "%s/%s/issues/%s",
        "pull": "%s/%s/pull-requests/%s",
        "commit": "%s/%s/commits/commit/%s",
        "compare": "%s/%s/branches/commits/%s..%s#diff",
        "hash_size": 7
    },
    "github": {
        "provider": "GitHub",
        "url": "https://github.com",
        "user_pattern": RE_GITHUB_USER,
        "issue": "%s/%s/issues/%s",
        "pull": "%s/%s/pull/%s",
        "commit": "%s/%s/commit/%s",
        "compare": "%s/%s/compare/%s...%s",
        "hash_size": 7
    }
}
```

_Note: I'm ignoring Twitter here because I haven't really put any thought into that._

Honestly, you might even be able to keep the `PROVIDER_INFO` the way it is. Whenever a new provider is defined you can just clone the `provider` it's based off of, replace anything that has been overridden (the `url` and `provider` properties come to mind), then loop over each of the properties replacing the original `url` part with whatever `url` the user provided. In this way you could even extend non-default providers.

### `RE_REPO_LINK_EXPRESSION`

Certainly the compiled `RE_REPO_LINK_EXPRESSION` would have to change to support additional URLs. Probably just build the string up on initialization and then compile it once all of the providers have been processed.

## Comments

I started to do this myself to make a PR but it turned out to be a bit more work than I realized it would be. If this is something you'd be interested in implementing let me know. I've at least thought about what the design would look like a bit.

facelessuser:
Closes #583

### DIFF
--- a/pymdownx/magiclink.py
+++ b/pymdownx/magiclink.py
@@ -35,6 +35,7 @@ from markdown.inlinepatterns import LinkInlineProcessor, InlineProcessor
 MAGIC_LINK = 1
 MAGIC_AUTO_LINK = 2
 
+RE_VALID_REPO_NAME = re.compile(r'[-a-z0-9]+', re.I)
 
 # Bare link/email detection
 RE_MAIL = r'''(?xi)
@@ -838,9 +839,17 @@ class MagiclinkExtension(Extension):
             host = custom['host']
             repo_type = custom['type']
             name = custom['name']
+            if RE_VALID_REPO_NAME.match(name) is None:
+                raise ValueError(
+                    "Custom repository names must only use characters [-a-zA-Z0-9_], '{}' is invalid".format(name)
+                )
             if name not in self.provider_info:
+                if repo_type not in ('github', 'gitlab', 'bitbucket'):
+                    raise ValueError("Custom repository type of '{}' is not supported".format(repo_type))
                 self.provider_info[name] = copy.deepcopy(PROVIDER_INFO[repo_type])
                 self.provider_info[name]['host'] = host
+            else:
+                raise ValueError("The custom name '{}' is already registered".format(name))
 
         github_hosts = []
         bitbucket_hosts = []
@@ -854,10 +863,10 @@ class MagiclinkExtension(Extension):
             if provider['type'] == 'github':
                 github_names.append(key)
                 github_hosts.append(provider['host'])
-            if provider['type'] == 'bitbucket':
+            elif provider['type'] == 'bitbucket':
                 bitbucket_names.append(key)
                 bitbucket_hosts.append(provider['host'])
-            if provider['type'] == 'gitlab':
+            elif provider['type'] == 'gitlab':
                 gitlab_names.append(key)
                 gitlab_hosts.append(provider['host'])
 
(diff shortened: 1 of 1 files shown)


facelessuser:
This is looking like it isn't as simple as changing the host for enterprise vs public available version. I'm not sure how much effort I am willing to put into this right now.

@gir-bot add S: needs-decision

facelessuser:
I'm going to close this for now as I am not confident this is flexible enough. I would need to be more sure before I release anything like this.

PR #701, opened 2019-10-19
Try a new strategy to drop appveyor

facelessuser:
We must use linux to use codecov's action, so try and cache coverage
data, and then combine and upload in separate step.


facelessuser:
Finally got it to upload the combined coverage, but it was accounted in  the report 🤷‍♂ . This has been exhausting. I wish Codecov had an action that worked on Windows.

facelessuser:
Okay, I was uploading the wrong format which is why it didn't work. XML is apparently the required format. Also, merging Linux and Windows is a bad idea. So we'll probably cover the data to XML, then archive that. Then when Windows tests are done, well use the Codecov action to upload each one. Hopefully that should work. If not, we might just keep Appveyor around until Codecov can provide an alternative.

facelessuser:
I spent too much time on this. I give up. The options just aren't there yet in GitHub CI, at least not for coverage and Python (when you want external pull requests to work).

PR #773, opened 2019-12-27
Deprecate extrarawhtml in favor of md_in_html in Python Markdown 3.2

facelessuser:


facelessuser:
Need to figure out how to run our document builder, that is currently restricted to <6.3 of this library, and it get it to build with 6.2. I thought you'd would handle that, but it keeps loading out local library despite installing 6.2.

facelessuser:
May have to resort to an earlier version of mkdocs-materia until it is updated to support 6.3.

facelessuser:
@gir-bot lgtm

PR #774, opened 2019-12-29
Drop Python 2.7

### LINKED ISSUE #705: Drop Python 2.7 support
Python Markdown will be dropping Python 2.7 in their next major release. We will probably follow as well. Maybe not next version, but if not, probably the following .

facelessuser:
Closes #705


facelessuser:
@gir-bot lgtm

Issue #794, opened 2020-01-17
MagicLink: Single-word hostnames not recognised as links.


willstott101:
We tend to give intranet servers a single-part hostname such as george so URLs end up being like this: https://george/wiki/

Links with hostnames like this are recognised and hyperlinked by GitLab and GitHub markdown. However this implementation does not recognise them as links.

I noticed this whilst using your sublime extension. When switching to an API back-end this was an obvious difference for me. I'd love to be able to use this when preparing docs for our local GitLab instance, without having to make network requests.

Thanks for your work btw, lots of small commits to format a doc correctly drives me crazy :)

facelessuser:
@gir-bot remove S: triage
@gir-bot add T: bug, S: confirmed

PR #795, opened 2020-01-17
MagicLink: Support single-item hostnames without a TLD

willstott101:
Closes: #794

### DIFF
--- a/pymdownx/magiclink.py
+++ b/pymdownx/magiclink.py
@@ -48,7 +48,7 @@ RE_MAIL = r'''(?xi)
 RE_LINK = r'''(?xi)
 (?P<link>
     (?:(?<=\b)|(?<=_))(?:
-        (?:ht|f)tps?://(?:(?:[^_\W][-\w]*(?:\.[-\w.]+)+)|localhost)|  # (http|ftp)://
+        (?:ht|f)tps?://(?:[^_\W][-\w]*(?:[^ \d][-\w.]+)+)|  # (http|ftp)://
         (?P<www>w{3}\.)[^_\W][-\w]*(?:\.[-\w.]+)+                     # www.
     )
     /?[-\w.?,!'(){}\[\]/+&@%$#=:"|~;]*                                # url path, fragments, and query stuff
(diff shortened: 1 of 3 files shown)


pymdownx/magiclink.py
```diff
@@ -48,7 +48,7 @@
 RE_LINK = r'''(?xi)
 (?P<link>
     (?:(?<=\b)|(?<=_))(?:
-        (?:ht|f)tps?://(?:(?:[^_\W][-\w]*(?:\.[-\w.]+)+)|localhost)|  # (http|ftp)://
+        (?:ht|f)tps?://(?:[^_\W][-\w]*(?:[^ \d][-\w.]+)+)|  # (http|ftp)://
```
facelessuser: This should probably apply to the `www.hostname` as well.

With that said, can you explain the approach of changing `\.` to `[^ \d]`?

facelessuser: Specifically why that pattern? Why are we avoiding `\d`?

facelessuser: Actually wondering if it would be better just to make the dot portion optional...

willstott101: Initially I went to `\.?` but that resulted in the negative test for `http://3628126748` failing (i.e. that became a link) so I went to `\D` but that resulted in `http://foo.bar/foo(bar)baz quux` being matched rather than excluding the `quux` being excluded. So I settled on `[^ \d]`.

I'm not 100% comfortable with that blacklist based on existing tests as an approach, but I'm not sure how else to prevent `http://3628126748` matching.

`www.hostname` does match with this regex, is that what we want?

willstott101: I unquoted the numeric link in my message, and it looks like GitHub allows that string. Looks like that test has been around since https://github.com/facelessuser/pymdown-extensions/commit/15e481e8bf757e8d8199db2c398e4bf2f9ce1439 I wonder if we should just let it be a URL?

facelessuser: No, I didn't mean `\.?`, I meant the entire dot portion of the pattern.  I went to code up the example, but then I needed to test it...so here is a pull: https://github.com/facelessuser/pymdown-extensions/pull/796.

As far as the failing test is concerned, I guess we could relax it a bit and allow numerical host name like GitHub and friends does.

facelessuser:
Marked as duplicate of #796 

@gir-bot remove S: needs-review
@gir-bot add S: duplicate

facelessuser:
@willstott101, Thanks for bringing this to my intention. I went ahead and merged #796 which approaches the pattern differently, but provides your desired functionality.

willstott101:
Yeah that looks good to me, thanks very much!

PR #796, opened 2020-01-17
Make dot portion of hostnames optional when auto-linking

### LINKED ISSUE #794: MagicLink: Single-word hostnames not recognised as links.

facelessuser:
Closes #795 
Closes #794


facelessuser:
@gir-bot  lgtm

PR #818, opened 2020-02-15
Use discord for chat

facelessuser:
Trying out Discord. No one is using the current Gitter chat, but my experience in other chat rooms with Gitter is not that great. The apps are kind of buggy on the different platforms I use. It isn't that instant with notifications. I just don't really like it...even if I wanted to.


facelessuser:
@gir-bot lgtm

PR #819, opened 2020-02-18
SuperFences/Highlight updates

### LINKED ISSUE #798: Adopt similar functionality as Python Markdown for classes and ids in code blocks
Reference issue: https://github.com/Python-Markdown/markdown/pull/816

@gir-bot add T: feature
@gir-bot lgtm

facelessuser:
- Add PHP markdown style insertion of classes and ids with attr lists
- Deprecate 'highlight_code' global option in SuperFences. It was true
  by default and disabling it doesn't make sense. It now does nothing
  and will be removed in the future.

Closes #798

### DIFF
--- a/docs/src/dictionary/en-custom.txt
+++ b/docs/src/dictionary/en-custom.txt
@@ -140,6 +140,7 @@ highlighter's
 hostnames
 html
 indepth
+injectable
 inline
 ish
 isn't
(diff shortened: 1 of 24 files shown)


facelessuser:
Reworked this a bit to ensure we have backwards compatibility. This shouldn't change any outputs except when you inject additional classes and ids. As far as `highlight_code` is concerned, it is still being removed.

facelessuser:
@gir-bot lgtm

PR #822, opened 2020-02-18
Allow overriding the base logic for inline and fenced code

### LINKED ISSUE #821: Provide a way to completely override the underlying highlighter for JS highlighters
Currently, SuperFences highlighter uses Pygments, or it does a generic output for JS highlighters.

You can add custom fences for specific keywords, but if you want to completely plugin and replace the highlighter to work better for a specific JS highlighter (let's say to better control JS highlighter specific options) it just isn't possible.

Currently, SuperFences and InlineHilite go through a list of formatters, and if none of the user ones match, it defaults to the general purpose one that they use to do basic highlighting.  But if we were able to override that base formatter, you could plug in any highlighter formatter you wanted.  I think we could do this simply by putting in logic that if a user specifies the custom name value as `*`, it will overwrite the base formatter. Then you could have your own Prism specific output and could control line numbers or whatever. 

@gir-bot add T: feature

facelessuser:
Closes #821

### DIFF
--- a/docs/src/dictionary/en-custom.txt
+++ b/docs/src/dictionary/en-custom.txt
@@ -142,6 +142,7 @@ html
 indepth
 injectable
 inline
+inlines
 ish
 isn't
 jQuery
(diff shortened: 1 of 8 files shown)


facelessuser:
@gir-bot lgtm

PR #857, opened 2020-03-20
Dark theme changes

facelessuser:


facelessuser:
@gir-bot lgtm

ofek:
Oh my! Please add this as an option to mkdocs-material

facelessuser:
Thanks! I automate it in other projects via a plugin: https://github.com/facelessuser/mkdocs_pymdownx_material_extras.  It's kind of meant as a way to patch my personal custom stuff into my projects.

You can see me using it here in this project: https://github.com/facelessuser/soupsieve/blob/master/mkdocs.yml#L103

I just set the custom color: https://github.com/facelessuser/soupsieve/blob/master/mkdocs.yml#L15

All the source is here if you want to adapt it for your own custom needs: https://github.com/facelessuser/pymdown-extensions/tree/master/docs/src/scss.

Eventually, Material will probably add a dark mode. Not sure what that will look like, but I didn't want to wait 🙂.

PR #863, opened 2020-03-24
UML is now demonstrated with Mermaid in the docs

facelessuser:
Mermaid is clearly a more feature rich UML tool. It can do all kinds of charts. Recent flowchart.js and js-sequence-diagrams, while working are becoming less compatible with each other. Sequences are dropping the library that used to be shared by both, but it also hasn't been updated in years...so.... It also doesn't have a reasonable CDN for it's recent release. Flowchart.js still works, but when you can get the kind of features out of Mermaid, it makes more sense to use that.

Nothing has changed that prevents the old UML implementation from working, we are just personally moving forward with Mermaid, and documented its usage now.


facelessuser:
@gir-bot lgtm

PR #899, opened 2020-04-18
Fix typo in a markdown kwarg name and fix import

felipelincoln:
In order to use `pymdownx.emoji.gemoji` it must have `emoji` being imported in the module `__init__.py` like this
```python
# pymdown-extensions/pymdownx/__init__.py
from . import emoji
...
```
But I don't think this is wanted so importing the submodule `emoji` directly would do the trick.

And as you can see in the `python-markdown` [reference](https://python-markdown.github.io/reference/#extension_configs), the correct name for the config kwarg is `extension_configs`.

### DIFF
--- a/docs/src/markdown/faq.md
+++ b/docs/src/markdown/faq.md
@@ -45,7 +45,7 @@ repository respectively. See [MagicLink](./extensions/magiclink.md) for more det
     see the related [FAQ question](#function-references-in-yaml) to see how to specify function references in YAML.
 
 ```py3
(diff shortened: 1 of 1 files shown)


facelessuser:
Yeah, I believe the intention was to use `import pymdownx.emoji`, which I clearly messed up on 🙂. Then you could specify `pymdownx.emoji.gemoji`. But using `from pymdownx import emoji` works as well. Thanks for the pull!

@gir-bot lgtm

PR #908, opened 2020-04-24
Inlinehilite: Add a default_lang config setting

scooby:
## Description

Adds a configuration option to inlinehilite to make code blocks without a specified language default to a given language.

### Updates

1. Adds the configuration option
2. Adds logic to handle defaulting
3. Adds a unit test
4. Adds the option to the inlinehilite docs

### Testing

I'm using it on a moderate sized document; also ran the inlinehilite unit tests to ensure no regression.

### Use case

I'm working on [a language specification](https://contravariance.gitlab.io/tenet-site/spec-expr-value.html) and I wind up with a ton of short snippets of code. It's a document about a single language and it's a custom lexer, so simply "use this language unless I say so" seems like the most expedient route.

If it's possible with `custom_inline`, I couldn't figure it out, and I'm using `mkdocs` so I'm stuck with whatever can be expressed in YAML.

### DIFF
--- a/docs/src/markdown/extensions/inlinehilite.md
+++ b/docs/src/markdown/extensions/inlinehilite.md
@@ -121,6 +121,7 @@ Option                    | Type         | Default       | Description
 `css_class`               | string       | `#!py3 ''`    | Class name is applied to the wrapper element of the code. If configured, this setting will override the `css_class` option of Highlight. If nothing is configured here or via or Highlight, the class `highlight` will be used.
 `style_plain_text`        | bool         | `#!py3 False` | When `guess_lang` is set to `#!py3 False`, InlineHilite will avoid applying classes to code blocks that do not explicitly set a language. If it is desired to have plain text styled like code, enable this to inject classes so that they can all be styled the same.
 `custom_inline`           | [dictionary] | `#!py3 []`    | Custom inline code blocks.
+`default_lang`            | string       | `#!py3 ''`    | Code blocks that do not explicitly set a language will default to this; useful for documents that primarily expect to use a single language.
 
 ---8<---
 links.txt
--- a/pymdownx/inlinehilite.py
+++ b/pymdownx/inlinehilite.py
@@ -107,6 +107,7 @@ class InlineHilitePattern(InlineProcessor):
         if not self.get_hl_settings:
             self.get_hl_settings = True
             self.style_plain_text = self.config['style_plain_text']
+            self.default_lang = self.config['default_lang']
 
             config = None
             self.highlighter = None
@@ -130,7 +131,7 @@ class InlineHilitePattern(InlineProcessor):
     def highlight_code(self, src, language, classname=None, md=None):
         """Syntax highlight the inline code block."""
 
-        process_text = self.style_plain_text or language or self.guess_lang
+        process_text = self.style_plain_text or language or self.default_lang or self.guess_lang
 
         if process_text:
             el = self.highlighter(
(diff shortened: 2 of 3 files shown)


facelessuser:
The whole point of me providing custom formatters is to avoid me supporting people's very specific use cases. Yours is not a common case, and I do not want to support lots of cases that only 1% of people use.

The example below would work fine. You would simply define a custom class that sets the language if none is provided, then override `*` which override the default handler.

```py
def _custom(src, language, class_name, md):
    """Test custom."""

    if not language:
        language = 'py3'

    return md.inlinePatterns['backtick'].highlight_code(src, language, class_name, md)

class TestInlineCustom(util.MdCase):
    """Test inline custom case."""

    extension = [
        'pymdownx.inlinehilite'
    ]

    extension_configs = {
        'pymdownx.inlinehilite': {
            'css_class': 'inlinehilite',
            'custom_inline': [
                {
                    'name': '*',
                    'class': 'highlight',
                    'format': _custom
                }
            ]
        }
    }

    def test_custom_specified(self):
        """Test when the language is specified."""

        self.check_markdown(
            '`#!css a{color: black;}`',
            '<p><code class="inlinehilite"><span class="nt">a</span><span class="p">{</span><span class="k">color</span><span class="p">:</span> <span class="kc">black</span><span class="p">;}</span></code></p>'
        )

    def test_custom_none(self):
        """Test when the language is not specified."""

        self.check_markdown(
            '`import module`',
            '<p><code class="inlinehilite"><span class="kn">import</span> <span class="nn">module</span></code></p>'
        )
```

So then you provide your own custom module and include it in your module:

```yml
markdown_extensions:
  - pymdownx.inlinehilite:
      custom_inline:
        - name: '*'
          class: highlight
          format: !!python/name:my_custom_module.my_custom_function
```

I do something similar that I've made available for the Material MkDocs theme to allow them to append custom emoji to the Emoji extension: https://github.com/facelessuser/mkdocs-material-extensions.

scooby:
Thanks for the detailed response! Didn't realize the way to use custom inline was that simple; I'm going to create an issue pointing to this PR for visibility.

facelessuser:
No problems! Thanks for creating a issue to guide others.

wxdrizzle:
Hello, thanks a lot for your solutions. Could you please provide some information about where I should put the `my_custom_module.py` file? Currently I just put it in my document folder (the folder containing `mkdocs.yml`), but when I executed `mkdocs serve`, an error appeared:
```
INFO     -  Building documentation...
Error: MkDocs encountered an error parsing the configuration file: while constructing a Python object
cannot find module 'my_custom_module' (No module named 'my_custom_module')
  in "/Users/drizzle/Documents/AICafe_documents/mkdocs.yml", line 85, column 23
```
I also tried to put it in the source code folder of mkdocs or pymdowns, but it still doesn't work.
A solution is to execute `python -m mkdocs serve` rather than `mkdocs serve`, but I'm also using [mike](https://github.com/jimporter/mike) to manage the version of my document, and it doesn't support `python -m xxx`, so I have to make `mkdocs serve` work.

wxdrizzle:
> Hello, thanks a lot for your solutions. Could you please provide some information about where I should put the `my_custom_module.py` file? Currently I just put it in my document folder (the folder containing `mkdocs.yml`), but when I executed `mkdocs serve`, an error appeared:
> 
> ```
> INFO     -  Building documentation...
> Error: MkDocs encountered an error parsing the configuration file: while constructing a Python object
> cannot find module 'my_custom_module' (No module named 'my_custom_module')
>   in "/Users/drizzle/Documents/AICafe_documents/mkdocs.yml", line 85, column 23
> ```
> 
> I also tried to put it in the source code folder of mkdocs or pymdowns, but it still doesn't work. A solution is to execute `python -m mkdocs serve` rather than `mkdocs serve`, but I'm also using [mike](https://github.com/jimporter/mike) to manage the version of my document, and it doesn't support `python -m xxx`, so I have to make `mkdocs serve` work.

I found the answer here: #928  Thank you!

PR #1008, opened 2020-07-03
Add SaneHeaders extension

### LINKED ISSUE #998: Add option in Magiclink to allow headers to not match issues #issue
The idea is to enable an option that can rewrite header handling to require a space after the `#`. This would allow things like `#2` to be recognized as an issue even if it is at the start of a line.

facelessuser:
Closes #998


facelessuser:
@gir-bot lgtm

PR #1009, opened 2020-07-05
Add full support attribute list style headers

facelessuser:
- Numerous changes related to how attribute lists are processed have been
implemented, some backwards incompatible. Some related to custom fences
and some related to how options are handled. Fallbacks for legacy custom
formatters and such have been removed.
- In the refactor, legacy
support for tabs in SuperFences has been removed. Tabbed should be
used instead.

This sort of mirrors the direction that Python Markdown's `fenced_code` extension is taking. Two big differences though is that we will not pass any keyword option to Pygments. We currently control what we wish to support. Second, we added an option in Highlight called `code_attr_on_pre` to control where attribute list style header options are attached in the generated structure.


facelessuser:
Unfortunately, backwards compatibility has not been preserved as far as custom fences is concerned. Some unhelpful functionality, such as attaching `linenums` to JavaScript style code blocks when Pygments is disabled, have been removed. Such features are actually not that helpful, and having full attribute list support will now allow users to properly configure JavaScript code blocks better.

Still need tests and need to update docs.

facelessuser:
We've made the transition easier, but there will still be some incompatible changes, but they are not as severe. We need to document things, but I think I'm done with this.

facelessuser:
@gir-bot lgtm

PR #1015, opened 2020-07-08
Remove extrarawhtml

facelessuser:


facelessuser:
With the changes coming in Python Markdown's HTML handling, the original extension is going to be rewritten it seems. So our old deprecated version may be obsolete soon anyway. Might as well go ahead and remove it now.

facelessuser:
@gir-bot lgtm

PR #1016, opened 2020-07-13
Fix Fatal Unpacking Bug

GiovanH:
With some combination of extensions and syntax, the stash `get` can fail here. Because the default is None, and None is not unpackable, this escalates and crashes the entire suite. This fixes that.

### DIFF
--- a/pymdownx/superfences.py
+++ b/pymdownx/superfences.py
@@ -847,7 +847,7 @@ class SuperFencesBlockPreprocessor(Preprocessor):
                 key = m.group(2)
                 indent_level = len(m.group(1))
                 original = None
-                original, pos = self.extension.stash.get(key)
+                original, pos = self.extension.stash.get(key, (None, None))
                 if original is not None:
                     code = self.reindent(original, pos, indent_level)
                     new_lines.extend(code)


GiovanH:
Style edits are welcome. There are other ways to do this check, of course. You can delay unpacking by assigning an object and checking to see if it's a tuple. This is just a concise patch.

facelessuser:
@GiovanH Thanks for the pull! I've never actually run into this issue, so I never caught this bug. Thanks for not only filing the issue, but also providing a fix.

@gir-bot lgtm

GiovanH:
👍
If you're curious, this is the extension (I wrote) that was creating some sort of conflict: https://github.com/GiovanH/peliplugins/blob/master/spoilerbox.py

It's definitely a weird edge case of some sort, and once your error handling was working again everything works fine.

facelessuser:
Curious. I know SuperFences has to implement a stash system to restore content in case it gets overzealous.

Since it runs as a preprocessor, it can't tell if a potential code block is under an HTML tag or under some indented code block and shouldn't get rendered. So in a postprocess step, it restores any that are left over that were incorrectly considered. I imagine it is somehow seeing your stashed ids and trying to find them in its own stash, but since they aren't there, it fails. I'm not 100% sure as I didn't look to close and have no reproducible examples, but if everything is working well for you now, then I guess there is nothing to worry about.

PR #1033, opened 2020-07-25
Document updates to support Material 5.5

facelessuser:
- Source links were deprecated, so do our own


facelessuser:
@gir-bot lgtm

PR #1041, opened 2020-08-02
Fix tabbed regression

facelessuser:
Broke some corner cases related to tab parsing with recent list fix


facelessuser:
@gir-bot lgtm

PR #1057, opened 2020-08-18
Remove tabbed fences feature mention

mlclmj:
It looks like this feature was retired in v8.0 but the feature mention wasn't removed from the docs


facelessuser:
Good catch!

facelessuser:
@gir-bot lgtm

PR #1165, opened 2020-12-03
Update Magiclink with recommendation for SaneHeaders

### LINKED ISSUE #1164: Magiclinks and issue tags need sane headers
Just a quick FYI, might be worth a document update to say you need to include sane headers when using the magiclinks plugin with `#` issues.  Without it the pull requests (using `!1`), work but `#1` gets picked up as an `<h1>`.  

I know this is pretty obvious but it took me a while to figure out which extension was causing the issue, so I think a quick note might help here.

Cheers for the extensions, they are awesome!

facelessuser:
Closes #1164


facelessuser:
@gir-bot lgtm

PR #1168, opened 2020-12-05
Add note about custom fences and logging

### LINKED ISSUE #1136: Don't swallow exceptions in custom SuperFence source formatter
When using the SuperFences extension in one of my projects, I noticed that exceptions raised inside a custom formatter are apparently caught — at least they never seem to reach stdout/stderr. Consider e.g. the following trivial source formatter:

```py
def source_formatter(source, language, css_class, options, md, **kwargs):
  print('[START]')  
  raise Exception('Boom!') ######## Raised but swallowed
  print('[END]')
  return f'<code>todo</code>'
```

Raising the exception appears to terminate `source_formatter` (`[END]` is never printed), but not the overall Markdown processing, which makes me suspect that the exception is either caught by SuperFences or by Python-Markdown itself.

**Question:** How can I prevent SuperFences (or Python-Markdown) from catching exceptions?

-------

Here is the minimal working example:

    import markdown
    import pymdownx.superfences

    def source_validator(language, inputs, options, attrs, md):
      return True

    def source_formatter(source, language, css_class, options, md, **kwargs):
      print('[START]')  
      raise Exception('Boom!') ######## Raised but swallowed
      print('[END]')
      return f'<code>todo</code>'

    superfences_config = {
      'custom_fences': [
        {
          'name': 'foo',
          'class': 'highlight',
          'format': source_formatter,
          'validator': source_validator
        }
      ]
    }

    source = '''
      ```foo
      x := y
      ```
    '''

    html = markdown.markdown(
      source, 
      extensions=['pymdownx.superfences'], 
      extension_configs={'pymdownx.superfences': superfences_config})

    print(html)

facelessuser:
Closes #1136

### DIFF
--- a/docs/src/markdown/extensions/superfences.md
+++ b/docs/src/markdown/extensions/superfences.md
@@ -397,6 +397,10 @@ Keys        | Description
 `format`    | A function that formats the HTML output. The function should return a string as HTML.
 `validator` | An optional parameter that is used to provide a function to validate custom fence parameters.
 
+!!! warning "Logging"
+    When a custom fence fails, the error will be swallowed up and the error will be handled gracefully. If logging is
+    desired, custom logging should be added into the custom function. SuperFences will not provide any.
+
 !!! new "New in 7.0"
     Starting in 7.0, you can override the base fence logic (the syntax highlighter) by specifying the custom fence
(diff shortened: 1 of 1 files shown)


facelessuser:
@gir-bot lgtm

PR #1186, opened 2020-12-24
Allow Markdown in Tabbed titles

### LINKED ISSUE #1063: Tabbed: Allow title content to be parsed for Markdown syntax
Tabbed uses a syntax very similar to Details and Admonitions. The title is within quotes, but the difference is that while this quoted content is parsed as Markdown for Details and Admonitions, it isn't for Tabbed. We should probably have Tabbed work the same way. While I wouldn't normally advise inserting anything there, there are some useful things, like emojis or something to make tabs a little more expressive.

facelessuser:
Closes #1063


facelessuser:
@gir-bot lgtm

PR #1191, opened 2021-01-01
Fix inconsistent internal code in relation to "guess_lang" in Highlight

### LINKED ISSUE #1190: What *is* the default for `guess_lang`?
1. https://github.com/facelessuser/pymdown-extensions/blob/6c3bad495b5b82b819121605364bbc2760dbae6e/pymdownx/highlight.py#L50-L51
2. https://github.com/facelessuser/pymdown-extensions/blob/6c3bad495b5b82b819121605364bbc2760dbae6e/pymdownx/highlight.py#L52
3. https://github.com/facelessuser/pymdown-extensions/blob/6c3bad495b5b82b819121605364bbc2760dbae6e/pymdownx/highlight.py#L194
4. https://github.com/facelessuser/pymdown-extensions/blob/29873da7921e7f2edf6c0e214ea90f501e1854b0/docs/src/markdown/extensions/highlight.md#L99

It seems that (1) is the dominating setting, and (4) documents it correctly,
but the description (2) is wrong, and the default (3) (which never really kicks in) might need to be changed to be less confusing.

facelessuser:
Closes #1190

### DIFF
--- a/pymdownx/highlight.py
+++ b/pymdownx/highlight.py
@@ -49,7 +49,7 @@ DEFAULT_CONFIG = {
     ],
     'guess_lang': [
         False,
-        "Automatic language detection - Default: True"
+        "Automatic language detection - Default: False"
     ],
     'css_class': [
         'highlight',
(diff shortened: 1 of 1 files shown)


facelessuser:
@gir-bot lgtm

PR #1200, opened 2021-01-12
Fixes corner cases where comment stripping is too greedy

### LINKED ISSUE #1199: Bug: Comments remover is "too hungry"
This code
```
<!-- END INCLUDE --></li>
Wrongly removed content
<!-- BEGIN INCLUDE -->

```
will end up removing all the content

If there is no new line, it works as expected. The problem lies (i think) in the lookahead rule. If it is not there, it works.

Used version: pymdown-extensions             8.0.1

I have used the RE from master branch, and the problem persist.

facelessuser:
Fixes #1199

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 8.1.2
+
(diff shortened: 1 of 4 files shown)


facelessuser:
@gir-bot lgtm

PR #1227, opened 2021-02-07
Fix add_attribute typo

gmjosack:
Function was named add_attriubtes. This corrects the typo but makes
an alias to maintain backwards compatibility with people using the
mispelling.

### DIFF
--- a/pymdownx/emoji.py
+++ b/pymdownx/emoji.py
@@ -54,13 +54,18 @@ Please update your custom index accordingly.
 """
 
 
-def add_attriubtes(options, attributes):
+def add_attributes(options, attributes):
     """Add additional attributes from options."""
 
     attr = options.get('attributes', {})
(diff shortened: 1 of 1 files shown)


facelessuser:
Good catch. Normally I'd say there is no need for the backwards compatibility as this is internals that I don't really support as an API...but I'm using it myself here: https://github.com/facelessuser/mkdocs-material-extensions/blob/master/materialx/emoji.py#L13. The funny thing is that I copied and pasted into that repo as well and didn't notice the misspelling 🤦.

Thanks!

@gir-bot lgtm

gmjosack:
Just noticed it while looking to make a wrapper for `to_png`. Also, love the bot :)

facelessuser:
> Also, love the bot :)

Ha, yeah, I see the resemblance :)

PR #1228, opened 2021-02-07
Update variable to be `def_non_std_image_path`

gmjosack:
When using `twemoji` I noticed if you pass through a name that's not unicode in the index you get unboundlocalerror for `def_non_std_image_path`. I believe everything in the current index is unicode so this likely never comes up but it seems to be a typo since `def_image_path` is defined twice consecutively.


gmjosack:
I only ran into this because I'm extending the twemoji index/generator. I'm not sure this will ever happen in practice but still seems like it was likely an oversight.

facelessuser:
Yup, looks like an oversight.

@gir-bot lgtm

PR #1319, opened 2021-04-29
Snippet enhancements

### LINKED ISSUE #1285: Add snippet to all files in a folder (snippet or abbr feature ?)
Hello, first of all thanks for your extensions, I just started looking into mkdocs and I found them really useful.
Having said that, I'd like to know if it's possible to include a snippet in all files in a folder, because I like to use it for abbr plugin but adding the snippet to each file is tedious and I may even forget to add it.

when I say folder, I mean maybe using a pattern, something like this

```
markdown_extensions:
  - abbr
  - pymdownx.snippets:
      base_path: doc/_snippets
      append:
         source: abbr.txt
         destination:
          - myfolder/*.md
          - another-folder/*.md
```

what I'm proposing is **append** configuration parameter

I understand that the problem is that a snippet can be placed everywhere but, at least in my case, I use it for links and abbr referenecs, so it really fits well the "append" concept, where you basically append the snippet at the end of each file

another approach I could maybe suggest is add an attribute to specify to append the snippet, so I can add the snppet syntax on the top of the document (wher I usually place meta data information since I'm using mkdocs-material <https://squidfunk.github.io/mkdocs-material/reference/meta-tags/> )

something like `--8<-- "refs.txt" [append] `

probably this concept would stick more to the abbr plugin but I suppose it make sense for snippets too?

I'm not into python and mkdocs so I can't spend time to do that myself, that's why I'm posting this otherwise I would have issued a PR :)
thanks

P.S.
I see it may be useful to your repo too because you manually added refs to many of your pages 😛

facelessuser:
- Add auto-append option to append snippets to every file
- Make base path option accept a list of base paths and even allow
  paths to specific files.

Fixes #1285


facelessuser:
@gir-bot lgtm

PR #1320, opened 2021-04-29
Fix bug with unescape order

### LINKED ISSUE #1290: Potential double-unescaping in pymdownx.highlight
Comparing this piece of code:

https://github.com/facelessuser/pymdown-extensions/blob/29873da7921e7f2edf6c0e214ea90f501e1854b0/pymdownx/highlight.py#L373-L378

with the bug that this pull request fixed:

https://github.com/Python-Markdown/markdown/pull/990/files#diff-934821cde110094c262f0f04b5e98feadce02b43ea311dd1c3d72d77b5c44860

-- I assume that the same issue affects this code and needs to be fixed in the same way.

I haven't run into this myself, though, and haven't confirmed the issue.

facelessuser:
& should be unescaped last to avoid issues with less than or greater
than

Fixes #1290

### DIFF
--- a/docs/src/dictionary/en-custom.txt
+++ b/docs/src/dictionary/en-custom.txt
@@ -233,6 +233,7 @@ tox
 uc
 un
 underscore's
+unescaping
 unicode
 unittest
 unstyled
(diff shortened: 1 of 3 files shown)


facelessuser:
@gir-bot lgtm

PR #1417, opened 2021-08-10
Feature/tabbed alternate

### LINKED ISSUE #1415: Draft: Revise HTML for Tabbed extension
__TL;DR: this is a draft to improve on the current implementation of the Tabbed extension. It is by no means ready for implementation. It should serve as a base for further discussion to learn whether we can gravitate towards a better solution than we currently have. There are still problems with this approach__.

## The problem

The Tabbed extension is pretty awesome and many users love it. It provides a lot of value, especially to Material for MkDocs. However, it suffers from some problems that are not solvable without additional JavaScript. The main problem is that on narrower screen sizes, tabs are broken onto separate lines, like here for example "Hide both":

Bildschirmfoto 2021-08-06 um 22 15 23

This cannot be mitigated, as the tabs markup is defined as follows:

``` html
<container>

  <!-- 1st tab -->
  <input 1 />
  <label 1 />
  <content 1 />

  <!-- 2nd tab -->
  <input 2 />
  <label 2 />
  <content 2 />

  ...
</container>
```

In order to make the labels overflow and scrollable, they would need to be located inside a container. With my knowledge of HTML and CSS, I'm very certain that there's no way to solve this problem without changing the underlying HTML. Furthermore, this would break the current tabs activation approach which relies entirely on co-location of `input` and `label` elements.

## A solution attempt

I've re-architected the tabs HTML and found a solution that allows us to overflow the tabs label container. It looks like this:

Bildschirmfoto 2021-08-06 um 22 22 05

https://user-images.githubusercontent.com/932156/128594641-00fb9e53-facc-43f2-8a2f-79c418baad27.mp4

The HTML can be found here, so just drop it into a `*.html` file and play with it:

<details>
<summary><b>Show revised HTML</b></summary>

``` html
<html>
  <head>
    <style>

      /* --- Boilerplate --- */

      body {
        font-family: sans-serif;
        margin: 40px;
      }

      body > label {
        display: inline-block;
        padding: 0px 0px 20px;
        cursor: pointer;
      }

      article {
        box-shadow: 0 0 10px black;
        overflow: hidden;
      }

      :checked ~ article {
        width: 400px;
      }

      /* --- Tab label --- */

      /* Tab input */
      .tabbed-set input {
        position: absolute;
        width: 0;
        height: 0;
        opacity: 0;
      }

      /* Tab label container */
      .tabbed-labels {
        display: flex;
        overflow: auto;
        box-shadow: 0 -1px 0 #CCC inset;
        scroll-snap-type: x proximity;
      }

      /* Tab label */
      .tabbed-labels > * {
        display: inline-block;
        padding: 10px;
        white-space: nowrap;
        scroll-snap-align: start;
        cursor: pointer;
      }

      /* Hide scrollbar for Chrome, Safari and Opera */
      .tabbed-labels::-webkit-scrollbar {
        display: none;
      }

      /* Hide scrollbar for IE, Edge and Firefox */
      .tabbed-labels {
        -ms-overflow-style: none;  /* IE and Edge */
        scrollbar-width: none;  /* Firefox */
      }

      /* Tab label is active */
      .tabbed-set input:nth-child(1):checked ~ .tabbed-labels > :nth-child(1),
      .tabbed-set input:nth-child(2):checked ~ .tabbed-labels > :nth-child(2),
      .tabbed-set input:nth-child(3):checked ~ .tabbed-labels > :nth-child(3) {
        color: deeppink;
        box-shadow: 0 -4px 0 deeppink inset;
      }

      /* --- Tab content --- */

      /* Inactive tab */
      .tabbed-content div {
        display: none;
      }

      /* Active tab */
      .tabbed-set input:nth-child(1):checked ~ .tabbed-content > :nth-child(1),
      .tabbed-set input:nth-child(2):checked ~ .tabbed-content > :nth-child(2),
      .tabbed-set input:nth-child(3):checked ~ .tabbed-content > :nth-child(3) {
        display: block;
        padding: 10px;
      }
    </style>
  </head>
  <body>

    <!-- Just for the demo -->
    <input type="checkbox" id="limit" checked>
    <label for="limit">Limit width of container</label>

    <!-- The interesting stuff -->
    <article>
      <div class="tabbed-set" data-tabs="10:3">
        <input id="__tabbed_10_1" name="__tabbed_10" type="radio" checked>
        <input id="__tabbed_10_2" name="__tabbed_10" type="radio">
        <input id="__tabbed_10_3" name="__tabbed_10" type="radio">
        <div class="tabbed-labels">
          <label for="__tabbed_10_1">A very long title</label>
          <label for="__tabbed_10_2">Another very long title</label>
          <label for="__tabbed_10_3">Well, why can't you just use shorter titles?</label>
        </div>
        <div class="tabbed-content">
          <div><p>Some stuff</p></div> 
          <div><p>More stuff</p></div> 
          <div><p>Even more stuff</p></div> 
        </div>
      </div>
    </article>
    <script>
      const tabs = document.querySelectorAll(".tabbed-set > input")
      for (const tab of tabs) {
        tab.addEventListener("change", () => {
          const label = document.querySelector(`label[for=${tab.id}]`)
          label.scrollIntoView({ behavior: "smooth" })
        })
      }
    </script>
  </body>
</html>
```

</details>

Here's how it works:

1. Labels and contents are grouped inside separate containers, allowing for better control of what is shown
2. `input` + `label` elements are targetted with `:nth-child(...)` selectors. While this seems unnecessarily bloated it is, AFAIK, the only viable approach. Furthermore, when transferred over the wire, it should compress very well, because the markup is largely the same, so `gzip` and friends should work very efficiently. We can provide, let's say, 10 selectors. That should be sufficient for 99.9% of all use cases. The CSS can be extended if more is necessary.

## Constraints

1. The print view of Material for MkDocs currently just expands all tabs and renders them below each other. A possible solution for this could be to add the tabs' title as an attribute to the content element and use a pseudo-element to render a label before the content container. I haven't tested this thoroughly, but it might be a start:

    HTML:

    ``` html
    <div class="tabbed-content">
      <div data-title="A very long title"><p>Some stuff</p></div> 
      ...
    </div>
    ```

    CSS:

    ``` css
    @media print {
      .tabbed-content > ::before {
        content: attr(data-title);
        ...
      }
    }
    ```

    Downside: some duplicate HTML (i.e. the tabs title), but personally, I could live with that.

2. We definitely shouldn't start relying on JavaScript for this functionality. The current CSS-only solution is pretty awesome, because it works on slow connections and when the site is partly ready due to missing JavaScript. For a comparison, try [Docusaurus's tabs implementation](https://docusaurus.io/docs/installation#running-the-development-server) after disabling JavaScript - it doesn't work at all.

3. We might need to keep the old implementation for older clients for some time.

---

Any feedback is greatly appreciated. If you don't wish to go down this path for whatever reason, feel free to close this issue. I've had this on my mind for a long time and wanted to give it a shot and publish it to get some feedback and ideas. I don't expect this to land quickly, since there is some stuff to work out.

Also, if GitHub issues is not the right place to discuss, feel free to move this to a discussion.

facelessuser:
This is experimental and still under review. This may or may not get merged.

Closes #1415


facelessuser:
@gir-bot lgtm

PR #1426, opened 2021-08-23
Experimental: Highlight Enhancements

### LINKED ISSUE #1138: Pygments' HtmlFormatter extra options
Hi @facelessuser and thanks for the fantastic `pymdown-extensions` package!

I'm looking to use extra Pygments' options like `linespans`, `lineanchors`, `anchorlinenos` and `nobackground` from the `HtmlFormatter`: https://pygments.org/docs/formatters/#HtmlFormatter .

`python-markdown`'s `code_hilite` supports them well: https://python-markdown.github.io/extensions/code_hilite/#usage

Is there any plan to support these extra-options or should I switch to `code_hilite`?

facelessuser:
Closes #1138


facelessuser:
@squidfunk Prototype of filename/title feature. I've styled and enabled it by default in this branch's documentation so it can easily be viewed. I will most likely not have it enabled by default when I launch.

Currently, users can specify a title and such by using `filename="some title"`. If desired, users can turn on auto-filename (via the `auto_filename` option in Highlight) which will grab the friendly name from the Pygments lexer and title case it for a suitable title. At any time, this can be overridden with `filename="some title"`.

Some Lexer names may not be preferable, so I have added an option called `auto_filename_mapping` which allows a user to override lexer names via a dictionary. This is done in this pull as an example.

The styling I've provided is just a placeholder. I have no idea exactly how Material will style this, but I thought this looked pretty good for the prototype:

Screen Shot 2021-08-23 at 9 58 42 AM

Could this also provide some sort of template syntax to insert the language? Maybe `filename="{lang}: some title"`. I'm not sure if I want to add something like that, but something like that could be possible. Could we have an option just to prepend language automatically? Sure, but then you could not override the language, so if something magic like this were wanted, the template route may be the way to go to pick up on user language overrides.

This is all experimental and subject to change. The approach and such are open to discussion.

facelessuser:
@squidfunk, I will most likely run with the `filename` option. I find it to be a pretty satisfying approach.

While I'm not against Material ignoring `filename` and going about a new approach using `data-attr`s, I see no reason to go down this path. I think it can provide pretty sane titles automatically, and users can provide their own mapping to override some of these automatic titles for ones they feel are better. A user can also always override a title per block as well. There may be room for improvements moving forward, so I'll at least mark it experimental for some time.

With that said, I'll see if I am able to look into `data-atttr` as well to add to Pygments code blocks. If not in this next release, the one after. I still think there may be uses for `data-attr` outside of code headers.

facelessuser:
@squidfunk So, I'm in the process of adding `data-attr` support. Right now, it will just attach any user specified `data-` attribute to the Pygments wrapper `<div>`. When not using Pygments, it always attaches the additional attributes to the `code` element or on the `pre` element if `code_attr_on_pre` is enabled. This is mainly because JavaScript highlighters usually require this. Pygments obviously doesn't require things to be on the `code` or `pre` element, and with Pygment's table format and the ability to use the legacy style which just used `<pre>` instead of `<pre><code>` it just seems easiest to attach them to `<div>`.

With all of that said, do you agree that attaching them to `<div>` is best? Or would you prefer that we try to attach them to either `<code>` or  `<pre>` depending on what is available and what the user has specified?

facelessuser:
Basic support for `data-` attributes and IDs for Pygments code blocks are now on the branch. For now, we are planning to attach such attributes to the Pygments' `<div>` wrapper. We can discuss things if this is undesirable, though I am shooting to get this out the door sooner rather than later. I still have more tests to write and documentation cleanup etc.

squidfunk:
> So, I'm in the process of adding data-attr support. Right now, it will just attach any user specified data- attribute to the Pygments wrapper `<div>` ... With all of that said, do you agree that attaching them to `<div>` is best? Or would you prefer that we try to attach them to either `<code>` or `<pre>` depending on what is available and what the user has specified?

From the top of my head, I'd say the outermost layer, which would be the wrapper div for Pygments, should be the best choice. Adding support for non-Pygments cases should probably also be possible in Material. However, I haven't had the chance to experiment with it, as the new search is currently eating up all my time. Could you open an issue over at Material, so we have a place to track support? I think we can do it similar to how we did it with tabs (which we still need to finish).

facelessuser:
All that is left is some tests. Will probably try to wrap up these highlight related features in the next couple of days.

facelessuser:
@gir-bot lgtm

facelessuser:
It should be noted that tables with line numbers do render a bit different, which makes sense. They do not have a div wrapper like I thought, so Pygments injects the `filename` into `<th>` row  that spans all columns.

PR #1428, opened 2021-08-24
Update dependencies and use node 16

facelessuser:


facelessuser:
From what I can tell, things look good. We were kind of forced to update to Node 16 if we wanted to stay on the latest and greatest versions of packages. VinylPaths forced us to run as an ESM module, but in the end, it all seems to be working. While annoying to have to debug all the new differences in this transition, it is done and working.

@gir-bot lgtm

PR #1452, opened 2021-09-19
Update doc deps (node) and allow scripts to work with instant mode

facelessuser:
Enable instant mode


facelessuser:
@gir-bot lgtm

PR #1465, opened 2021-09-23
Experimental: Embed videos

### LINKED ISSUE #896: [Feature Request] YouTube Embeds
I currently do some very ugly hacks in JavaScript at the moment to get a few missing Markdown features working.

One of them is a tag for YouTube embeds. Currently I'm scanning the `alt` tag of images for `YOUTUBE` and then replace the image with an iframe like so:

```js
const elements = document.querySelectorAll("img[alt=\"YOUTUBE\"]")

  Array.prototype.forEach.call(elements, (el, i) => {
    const id = el.getAttribute("title").split("/")[el.getAttribute("title").split("/").length - 1]
    const oldClass = el.getAttribute("class")
    const iframe = document.createElement("iframe")

    iframe.title = "YouTube"
    iframe.src = `https://www.youtube.com/embed/${id}?modestbranding=1&amp;`

    if (oldClass !== "null") {
      iframe.className = `video-embed ${oldClass}`
    } else {
      iframe.className = "video-embed"
    }

    el.replaceWith(iframe)
  })
```

Now if we could do this natively and on compile instead of run, that would be great.

### LINKED ISSUE #897: [Feature Request] Video Embeds
Since proper Markdown support is not going to happen anytime soon, would it be possible to implement it in material-extensions like suggested here? https://talk.commonmark.org/t/embedded-audio-and-video/441/102

There is https://github.com/cmrd-senya/markdown-it-html5-embed as an example implementation.

facelessuser:
## Embedding Video/Audio File Types

Resolves #897 

This is an experimental branch that explores syntax to embed media. The idea is to use the existing `` syntax that has been historically used for images, and instead, use it for media in general.

```
![](mov.webm)
```

"Img" links are processed as normal, and once `attr_list` has processed them, we intercept them in a Treeprocessor. There we identify any links that appear to be media links. Once we've found such links we wrap them in their appropriate element tag (`<video>` or `<audio>`). Inside the "media" tag, the video or audio file with be specified via a `<source>` element.

We do have a default mapping for some known extension types, but some can be either audio or video, if our default assumption is incorrect, the user should specify the `type` attribute using `attr_list`.

Since these are processed after `attr_list`, this means we can apply whatever attributes we want to the element(s) and leverage those. When we are processing the attributes attached to the given tag, we can sort them based on whether they are appropriate for the `<source>` (or `<track>` -- subtitles) element or the parent container.

For instance, here we set `preload` to `metadata` for the media container.

```
![](mov.webm){preload=metadata}
```

But here, since `webm` can be ambiguous, and we default to assuming `video`, we can specify on the `<source>` element that the type is `audio/webm`. Since we are aware of which attributes are `<source>` specific, we can specify them together and it will all get sorted out.

```
![](mov.webm){type="audio/webm" preload=metadata}
```

Anything non-specific to either element will be attached at the parent container level.

If desired, a user can specify globally, via options, a set of default attributes to apply on every media object. If a user explicitly sets that attribute on a media object, the global value will of course be overridden.

For all videos, we will specify a fallback, the last resort for when the video or audio isn't handled. It is a simple message that allows the user to try and download the file directly. 

```html
<video controls preload="metadata">
<source src="https://dl8.webmfiles.org/elephants-dream.webm" type="video/webm">
<a download="" href="https://dl8.webmfiles.org/elephants-dream.webm">https://dl8.webmfiles.org/elephants-dream.webm</a>
</video>
```

If the user provides an `alt`, we will use that to provide a description -- in the case that no `alt` is specified and a title is, title can double as an alt. Assuming none of these are provided, the fallback will simply be the source link.

```html
<video controls preload="metadata"><source src="https://dl8.webmfiles.org/elephants-dream.webm" type="video/webm"><source src="./images/elephants-dream.webm" type="video/webm">
<a download="" href="https://dl8.webmfiles.org/elephants-dream.webm">Elephant has a dream</a>
</video>
```

Lastly, we provide a mechanism for fallback sources. If multiple, same types of media (audio or video) are grouped together, the first one will be assumed as the main video, and all others will be assumed fallbacks if they specify the `fallback` attribute. This can be used to specify multiple video types in case a browser does not support one type. Additionally, `.vtt` sources can be used and will be included in whatever the parent is, audio or video -- `.vtt` sources are inserted as `<track>` elements, not `<source>`.

```
![A movie about stuff](mov.webm)
![](mov.mp4){fallback}
```

```html
<video controls= preload="metadata">
<source src="mov.webm" type="video/webm">
<source src="mov.mp4" type="video/mp4">
<a download="" href="mov.webm">mov.webm</a>
</video>
```

Source-specific attributes will apply to each source, but the "media" and general attributes will be taken only from the main source.

```
![Audio about stuff](audio.mp3){loop=''}
![](audio.webm){fallback type="audio/webm"}
```

```html
<audio controls loop preload="metadata">
<source src="audio.mp3" type="audio/mpeg">
<source src="audio.webm" type="audio/webm">
<a download="" href="audio.mp3">Audio about stuff</a>
</audio>
```

## Embedding Media Services

Resolves #896 

Embedding video services such as Youtube uses the same syntax as the above `some alt text`. These media links will be converted to an appropriate embedded link and wrapped in an `<iframe>`.

`<iframe>` have no concept of `alt`, but it is generally encouraged that a `title` is applied. So in this case, `alt` will double as a `title` if `title` is not defined. If neither `alt` or `title` is defined, something generic like `YouTube video player` will be provided via the default configuration.

```
![Random YouTube link](https://www.youtube.com/embed/ZTjzHsAU5-U)
```

```html
<iframe allowfullscreen="" frameborder="0" src="https://www.youtube.com/embed/ZTjzHsAU5-U" title="Random YouTube link"></iframe>
```

Things are abstracted in such a way that someone could provide their own function for a specific video service, but as of right now, only: Vimeo, YouTube, and Dailymotion are supported out of the box.

Global default attributes can be defined per media service.

Feedback is welcome!

### DIFF
--- a/docs/src/dictionary/en-custom.txt
+++ b/docs/src/dictionary/en-custom.txt
@@ -20,6 +20,7 @@ CommonMark
 CriticMarkup
 Ctrl
 DOM
+Dailymotion
 Dicts
 Donath
 ElementTree
@@ -110,6 +111,7 @@ UTF
 Uml
 Unescape
 Validators
+Vimeo
 Virtualenv
 Waylan
 YAML
@@ -142,6 +144,7 @@ emojione
 escaper
 eslint
 facelessuser
+fallbacks
 formatter
 formatter's
 formatters
--- a/docs/src/mkdocs.yml
+++ b/docs/src/mkdocs.yml
@@ -153,6 +153,7 @@ markdown_extensions:
   - pymdownx.tabbed:
       alternate_style: true
   - pymdownx.saneheaders:
+  - pymdownx.embed:
 
 extra:
   social:
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -153,6 +153,7 @@ markdown_extensions:
   - pymdownx.tabbed:
       alternate_style: true
   - pymdownx.saneheaders:
+  - pymdownx.embed:
 
 extra:
   social:
--- a/pymdownx/embed.py
+++ b/pymdownx/embed.py
@@ -1,0 +1,186 @@
+"""Library for embedding media and such."""
+from markdown import Extension
+from markdown.treeprocessors import Treeprocessor
+from markdown import util as md_util
+import xml.etree.ElementTree as etree
+import re
+import copy
+
+
+class EmbedMediaTreeprocessor(Treeprocessor):
+    """
+    Find video links and parse them up.
+
+    We'll use the image link syntax to identify our links of interest.
+    """
+
+    # Current recognized file extensions
+    MIMES = re.compile(r'(?i).*?\.([a-z0-9]+)$')
+
+    # Default MIME types, but can be overridden
+    # These are just MIME types we know, but some can be
+    # audio or video, and we cannot predict without a type
+    # in those cases. So any of these can be overridden if
+    # a type is provided, but if none is provided, we take
+    # our best guess.
+    MIMEMAP = {
+        'mp4': 'video/mp4',
+        'webm': 'video/webm',
+        'mp3': 'audio/mpeg',
+        'ogg': 'audio/ogg',
+        'wav': 'audio/wav',
+        'flac': 'audio/flac'
+    }
+
+    def __init__(self, md, video_defaults, audio_defaults):
+        """Initialize."""
+
+        self.video_defaults = video_defaults
+        self.audio_defaults = audio_defaults
+
+        super().__init__(md)
+
+    def run(self, root):
+        """Shorten popular git repository links."""
+
+        # Grab the elements of interest and form a parent mapping
+        links = [e for e in root.iter() if e.tag.lower() in ('img',)]
+        parent_map = {c: p for p in root.iter() for c in p}
+
+        # Evaluate links
+        for link in reversed(links):
+
+            # Save the attributes as we will reuse them
+            attrib = copy.copy(link.attrib)
+
+            # See if source matches the audio or video mime type
+            src = attrib.get('src', '')
+            m = self.MIMES.match(src)
+            if m is None:
+                continue
+
+            # Use whatever audio/video type specified or construct our own
+            # Reject any other types
+            mime = m.group(1).lower()
+
+            # We don't know what case the attributes are in, so normalize them.
+            keys = set([k.lower() for k in attrib.keys()])
+
+            # Identify whether we are working with audio or video and save MIME type
+            mtype = ''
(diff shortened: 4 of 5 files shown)


facelessuser:
Maybe we should split this into a new extension? It doesn't necessarily rely on anything in MagicLink really. Might keep things cleaner and easier to maintain.

facelessuser:
I guess it is link related 🤷🏻. We can maybe at least break MagicLink into submodules and confine the media stuff to one submodule 🤔 .

facelessuser:
Hmm. Not every user may desire the default fallback to be in Engllish...maybe we should make the message configurable, or strip it down to only what is provided by `alt`.

facelessuser:
Move embedding logic to a new extension called Embed (need a better name?).

No more complex fallback message. Just show a download link with "alt" text or the URL. Most browsers support HTML5 video at this point. People will most likely never see this, and if they do, that's what they'll get.

flywire:
> Feedback is welcome!

If you are interested in captions this is very much a use I had envisaged for https://github.com/flywire/caption (assuming on a line by itself) but I never managed to apply it to user-specified content types. The unresolved issue in caption is overriding defaults.

facelessuser:
@flywire, I understand the idea behind this logic, but doing it this way will nest `<figure>` under `<p>` which isn't really proper. But the reasoning behind triggering off of title does make sense.

That is because `` is handled as inline content as `` is considered inline. `<figure>` should really be handled at the block level. I currently do something like this. But yeah, it is more verbose.

```html
Example audio

<figure markdown=1>

![](./images/file_example_MP3_700KB.mp3)

<figcaption markdown=1>This is a **song**</figcaption>
</figure>
```

In order to properly handle images, videos, and audios in captions, we'd either need some kind of figure container syntax or handle `` at the block level, which may be weird as then you can't really insert images as inline any more. 

At this time, I would probably pass on auto-captioning as I think this would need more thought. While what you suggest would work, it is probably not "proper". Granted, that may not matter to some, but it's just something to note.

BjoernSchotte:
As I'm interested in the embed feature, is there anything I can help with?

facelessuser:
I think for starters, getting some general feedback on the current implementation.

facelessuser:
I'm closing this issue PR for now. Moving forward, I think we'll be exploring such things with the new general-purpose blocks that are currently in beta.

PR #1472, opened 2021-09-29
Escape HTML entities properly

facelessuser:
The way we were doing them could lead to weird TOC entries and slugs


facelessuser:
@gir-bot lgtm

PR #1503, opened 2021-11-10
If linenums is globally true allow disabling linenums per block

### LINKED ISSUE #1486: Deactivate linenums on a per-codeblock basis, when linenums are set to True in mkdocs.yml
### Description

Would it be possible to add the following feature (to be used with mkdocs, and mkdocs material) :
1°) Activate automatic linenumbering for all codeblocks, by default, in the file `mkdocs.yml` with the syntax :
```
markdown_extensions:
  - pymdownx.highlight:
      linenums: True
```
AND
2°) Deactivate automatic linenumbering on a per-codeblock basis, for example with the following syntax in a specific codeblock:

    ```python linenums="0"
    print("That would be reeeaally great, and thank you sooo much!")
    ```

### Benefits

A new feature that might be really useful to some users :)

### Solution Idea

_No response_

facelessuser:
A code block in SuperFences can specifying the starting line as zero
to disable global linenums for that specific code block.

Resolves #1486

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -2,6 +2,8 @@
 
 ## 9.1
 
+- **NEW**: Highlight: If `linenums` is enabled globally via the `highlight` extension, and a code block specifies a line
+  number of zero (e.g. SuperFences), disable line numbers for that code block.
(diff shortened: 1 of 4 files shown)


facelessuser:
Needs targeted unit testing and documentation.

facelessuser:
@gir-bot lgtm

PR #1567, opened 2022-01-17
Fix Arithmatex escape issue

### LINKED ISSUE #1564: Arithmatex: removing dollar inline syntax gives "IndexError: no such group"
### Description

When configuring Arithmatex to disable the `$` syntax for inline mathematics with this configuration:
```yaml
# ...
  - pymdownx.arithmatex:
      # ...
      inline_syntax: [round]
```
Any markdown which contains `\( \)` (which is the`round` syntax left enabled) will fail to render with this exception:

```
Traceback (most recent call last):
  ...
  File ".../site-packages/markdown/core.py", line 268, in convert
    newRoot = treeprocessor.run(root)
  File ".../site-packages/markdown/treeprocessors.py", line 371, in run
    self.__handleInline(text), child
  File ".../site-packages/markdown/treeprocessors.py", line 131, in __handleInline
    data, matched, startIndex = self.__applyPattern(
  File ".../site-packages/markdown/treeprocessors.py", line 272, in __applyPattern
    node, start, end = pattern.handleMatch(match, data)
  File ".../site-packages/pymdownx/arithmatex.py", line 254, in handleMatch
    escapes = m.group(4)
IndexError: no such group
```

All of the following settings for `inline_syntax` still work:
```yaml
[round, dollar] # this is the default anyway
[dollar]
[]
```
So it's just this combination of round only that breaks.

This issue also appears to be unrelated to any of the other Arithmatex options (e.g. `smart_dollar: false` does not fix it)

### Minimal Reproduction

1. Configure pymdown-extensions to enable Arithmatex and set `inline_syntax: [round]`
2. Try to render the markdown `\( 1 + 2 = 3 \)`
3. Observe the error

### Version(s) & System Info

- Operating System: Arch Linux
- Python Version: 3.10.1
- Package Version: 9.1

facelessuser:
Fixes #1564

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 9.2
+
+- **FIX**: Arithmatex: fix issue where if you limit the inline syntax to a single input type, escape logic could throw
+  an error.
+
 ## 9.1
 
 - **NEW**: Highlight: If `linenums` is enabled globally via the `highlight` extension, and a code block specifies a line
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -2,8 +2,8 @@
 
 ## 9.2
 
-- **FIX**: Arithmatex: fix issue where if you limit the inline syntax to a single input type, escape logic could throw
-  an error.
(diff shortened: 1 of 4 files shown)


facelessuser:
@gir-bot lgtm

PR #1568, opened 2022-01-18
Add special exception to force a failure, if desired in custom fences

### LINKED ISSUE #1526: Allow custom superfences formatter to raise exceptions
### Description

I would like to be able to raise an exception from a custom fence formatter, like so:

```python
def fence_custom(source, language, class_name, options, md, **kwargs):
    raise Exception("hi")
```

Now, if an error occurs, it is ignored and input is formatted as a code block.

A [validator](https://facelessuser.github.io/pymdown-extensions/extensions/superfences/#validators) is not an option because I'm interested in validating the `source`.  

P.s. Really like your work on custom fence formatters & the documentation on your approach with mermaid really helped! 

### Benefits

I'm writing a new mkdocs plugin ([link](https://github.com/timvink/mkdocs-charts-plugin)), and this would help end users find errors their input code blocks instead of silently failing. 

### Solution Idea

Some ideas:

- Add a new validator for the source
- Add a boolean flag argument to the custom fence function to allow raising errors
- Add an option to the `pymdownx.superfences.custom_fences` to allow raising errors

facelessuser:
Resolves #1526

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -2,6 +2,9 @@
 
 ## 9.2
 
+- **NEW**: Custom fence validators and formatters can now be forced to raise an exception by raising a
+  `SuperFencesException`.
+- **FIX**: Fix case where custom fence in a blockquote was not gracefully handled.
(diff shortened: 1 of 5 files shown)


facelessuser:
@gir-bot lgtm

PR #1573, opened 2022-01-21
refactoring code with For Else

zjzh:
refactoring code with For Else which is more pythonic, concise and efficient; how do you think this change which has practical value？

### DIFF
--- a/tests/test_extensions/test_superfences.py
+++ b/tests/test_extensions/test_superfences.py
@@ -31,14 +31,13 @@ def custom_validater_exploder(language, inputs, options, attrs, md):
 
 def custom_validator(language, inputs, options, attrs, md):
     """Custom validator."""
-
     okay = True
     for k in inputs.keys():
         if k != 'opt':
             okay = False
             break
-    if okay:
-        if inputs['opt'] != "A":
+    else:
+        if inputs['opt'] != 'A':
             okay = False
         else:
             options['opt'] = inputs['opt']


facelessuser:
1. There is nothing un-pythonic about the current code.
2. The change you made actually changes the logic. It is not more efficient, it breaks the code.

We are only trying to evaluate `inputs` if it 1.) has items and 2.) the item key is named `opt`. Your change will evaluate `inputs` with the assumption that it found `opt` when it did not. This is not more efficient, this is simply wrong.

Take this example which demonstrates the behavior:

```
>>> l = []
>>> for x in l:
...     if x == 1:
...         break
... else:
...     print("okay")
... 
okay
```

This is not what we are aiming for.

While I do use `for/else` on occasions, it is usually less understandable to the average person, as this pull request demonstrates.   Though, I have not done a benchmark to prove whether it is in fact "more efficient", readability and understandability are very important, sometimes more important than efficiency if we are coming from a maintainability perspective, especially if the gain in efficiency is not substantial.

I appreciate your desire to help make the project better, unfortunately, these changes, at least this time, cannot be accepted.

@gir-bot add S: rejected

zjzh:
@facelessuser Thanks for your comments. However, the code should be correct. Let's look the following code, it will print okay. 
```
>>> l = []
>>> for x in l:

...     if x == 1:
...         break
... else:
...     print("okay")
... 
okay
```

However for the commits code, even though inputs is empty, the okay is True so it will execute the body of if okay. Therefore, we change if okay with else, the code still is correct.

```
okay = True
for k in inputs.keys():
        if k != 'opt':
            okay = False
            break
if okay:
        if inputs['opt'] != "A":
            okay = False
        else:
            options['opt'] = inputs['opt']
```

facelessuser:
Ah, then my original code is also wrong. We would not normally refactor it to be wrong in a more "efficient" way. That is a good catch though and is indeed a bug. What we'd like to do is probably shortcut out if we have no inputs.

```py
okay = True

# No inputs to evaluate
if not inputs:
    return okay

# Invalidate any bad keys
for k in inputs.keys():
    if k != 'opt':
        okay = False
        break

# Invalidate bad option values
if okay:
    if inputs['opt'] != "A":
        okay = False
    else:
        options['opt'] = inputs['opt']
```

My point though is simply that the for/else construct is a confusing thing for a lot of Python users,  which is why I generally avoid using it and it is often not seen in many other libs for similar reasons. That's not to say some people don't use it, but you really don't see it in many places. If I was in a time-critical place (which these tests are not), and this optimization is not just syntactic sugar and actually was more efficient (keep in mind I have not measured timing and optimizing without measuring is not useful), I would maybe consider such a use case.

With all of that said, the original code clearly has a bug. I think at the time it was coded with the assumption that none of the tests were going to pass in fences without options, which they didn't. So normally, gracefully handling such cases is the better way to go, so this would need to be fixed...except that I coincidentally have an issue open to test and handle such cases: https://github.com/facelessuser/pymdown-extensions/issues/1526. We are writing tests for these cases and I have [repurposed this failure in a test](https://github.com/facelessuser/pymdown-extensions/pull/1568/commits/651d3b8fb29db196e541eea9ce1f76da72e400a5) since you've pointed it out 🙂.

Generally, when custom fences fail with exceptions, we gracefully ignore the custom fence and move on. This is generally because we don't want to cause the whole document parsing to fail because of a faulty custom fence. But, some would like to force a failure no matter what. So, if the user forces a `SuperFencesException` this will halt all processing and cause Markdown termination to end abruptly. So, we will be re-using this bug, previously unintentional, but now intentional 🙂.

Anyways, I appreciate you finding this bug, and while this specific pull still won't be accepted for the above reasons, I look forward to other contributions if you happen to locate some improvements. Generally, I'm less keen on for/else refactors, but I'm sure there are plenty of other improvements that can be made.

zjzh:
@facelessuser Thank you. Actually the original code or new code you will change into the following code, which will still make programmers confuse the functionality. The confusing key is in the for loop. For example, let's assume that inputs['opt']=="A", when inputs have more than 1 keys such as 'opt' and 'a', the okay will be False. And when keys become 'opt' and 'z'. It will come into the if okay:.... and okay will be True. So the functionality of the code fragments especially make people confuse. Therefore, maybe you could check the functionality of the code fragment. 

> okay = True
> 
> # No inputs to evaluate
> if not inputs:
>     return okay
> 
> # Invalidate any bad keys
> for k in inputs.keys():
>     if k != 'opt':
>         okay = False
>         break
> 
> # Invalidate bad option values
> if okay:
>     if inputs['opt'] != "A":
>         okay = False
>     else:
>         options['opt'] = inputs['opt']

Actually, it could be refactored into the following code:

if 'opt' in inputs:
    if inputs['opt'] != "A":
        return False
   else:
        return True
else:
   options['opt'] = inputs['opt']

By the way, I understand that programmers have their own coding style, for the for k in inputs.keys(), it could be changed into for key in inputs. When you understand for/else, I believe that you will like it.

facelessuser:
Let's keep in mind this is a contrived, toy example created only for testing. We are simply testing the invalidation of a custom fence's header. That was the whole purpose of this bit of code, the actual options don't matter, only how it interacts with SuperFences.

> For example, let's assume that inputs['opt']=="A", when inputs have more than 1 keys such as 'opt' and 'a', the okay will be False.

Again, this is a contrived, toy example, but this is exactly the way it was coded, it seems you understand the logic fine. It seems what you are actually confused about is *why* I coded it that way, but that was more just to test the invalidation of options in arbitrary ways using the custom validator. That's exactly how it is supposed to work. It's not meant to be a real example that people go out and start using. It's a silly test example, but it tests what we need it o.

What we are really testing is what happens when the invalidator rejects things in the header of the custom fence. Does SuperFences choke or does it properly reject the fence and move on? We don't care if these specific options make sense, just that the validator responds as we expect and SuperFences does what it is supposed to. This is test code and **only** test code.

We were never testing the absence of options on this specific, contrived example, but your suggested refactor brought that point up.  I've conceded that, yes, if this was real production code, it should  handle no options in a sane manner as the current code would throw an exception. I thought we were handling it, but we were in fact not. Again, as this is only test code, and we were not testing that case, it was no big deal, so normally, I'd accept a fix for it, but It just so happens that I have an open pull request in which we can repurpose this normally bad behavior to test exception handling, so that is what I've done.

As for `for/else`, I generally don't like the syntax. I think I've used it once or twice in the past, but I generally stay away from it on purpose.

> the `if okay` could still be changed into `else`. And for the `for k in inputs.keys()`, it could be changed into `for key in inputs`.

We are now moving into a discussion that is based purely on stylistic preferences. Sure we *could* change it into an else, but we *could* also not. Yes, technically, we could just say `inputs` instead of the more explicit `inputs.keys()`, but now we are getting very nit-picky.

I'm all for fixing bugs, optimizing code, fixing typos and misspellings, etc. I'm even okay if a refactor makes things dramatically more understandable. I'm also willing to reject optimizations if they come at the cost of maintainability (what I perceive to be less understandable and/or fragile). I admit that the toy example is silly, but it does what it is meant to, we shouldn't overthink it too much 🙂. 

The rest of the arguments are based on personal coding preferences, which I'm afraid I'm just not willing to waste time on. If I allowed the code to change every time someone preferred coding things one way or another, the code would be in constant flux. To have a pull request accepted, I need changes that add real benefits: fixes, features, typo corrections, optimizations, etc.

I hope you can understand this position, even if we approach things differently 🙂.

PR #1585, opened 2022-01-29
Add power and fingerprint key to default key mapping

### LINKED ISSUE #1582: Addition of touch-id-sensor for keys extension
### Description

It looks like the [key map](https://facelessuser.github.io/pymdown-extensions/extensions/keys/#key-map-index) for the keys extension does not include an entry for the touch ID sensor on some Macbooks (and potentially other laptops). I noticed that there is also no entry for a power key.

### Benefits

It would be possible to refer to the key when writing documentation about how to easily lock your screen.

### Solution Idea

_No response_

facelessuser:
Resolves #1582


facelessuser:
@gir-bot lgtm

PR #1596, opened 2022-02-07
Add option to attach language classes to Pygments code blocks

### LINKED ISSUE #1572: Keep the specified language CSS class when colorizing a code block
### Description

For a code block like this,

    ```python
    a = True + 1
    ```

the HTML that `pymdownx.highlight` produces starts with the following:

`<div class="highlight"><pre><span>`

The code block is highlighted with the correct language and then the language specifier is not present anywhere in the output.

---

And note that the highlighted case actually differs from the "un-highlighted" use of `fenced_code_blocks`:

https://python-markdown.github.io/extensions/fenced_code_blocks/#language says:

> Either of the above examples will output the following HTML: `<pre><code class="language-html"`...

### Benefits

I think that keeping the language identifier in the output would let users make customizations more safely. For example, one could write this CSS to make "`$ `" terminal prompts non-selectable with mouse:

```css
.highlight.language-console .gp { /* Generic.Prompt */
  user-select: none;
}
```

([live example with this](https://mkdocstrings.github.io/crystal/))

But right now one is forced to write this unqualified:

```css
.highlight .gp { /* Generic.Prompt */
  user-select: none;
}
```

which also can affect blocks in other languages where it may not have been expected.

---

Another major use case might just be to provide a different look per language, say, a background color or an icon preceding it.

### Solution Idea

So I think `pymdownx.highlight` should be able to produce HTML that includes the language as a CSS class, for example:

`<div class="highlight language-console"><pre><span>`

---

Right now there's only a workaround to achieve such an output -- to write, for example,

    ```{.console .language-console}
    $ pip install pip
    ```

facelessuser:
Resolves #1572


facelessuser:
@gir-bot lgtm

PR #1612, opened 2022-02-19
Fix typo

pawamoy:
Unless I'm missing something?

### DIFF
--- a/docs/src/markdown/extensions/superfences.md
+++ b/docs/src/markdown/extensions/superfences.md
@@ -852,8 +852,8 @@ extension_configs = {
     "pymdownx.superfences": {
         "custom_fences": [
             {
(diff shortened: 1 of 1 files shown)


facelessuser:
I am using `diagram` in all my Mermaid diagrams so I don't accidentally trigger MkDocs Material's included Mermaid support. I had my own solution before Material introduced it and I didn't want it to break mine. In the future, I may move over to Material's, but right now I haven't.

Most likely I copied and pasted the results from my own config which uses `diagram` and forgot to change it to `mermaid` for the documentation.

pawamoy:
I see, thanks for the explanation!

facelessuser:
@gir-bot lgtm

Issue #862, opened 2020-03-23
Replace B64 with a general purpose image inline extension


facelessuser:
Will do everything that B64 does now, but will provide an syntax to force inlining, and optionally a mode to inline all images as B64 does now. Will handle PNG, GIF, JPEG, and additionally SVG.

Ref https://github.com/squidfunk/mkdocs-material/pull/1475

matoszz:
Stumbled across this as I've been searching for a good solution to have SVG's with functional links embedded into a mkdocs material setup...your docs note `support for PNG, JPEG, and GIF` but looking at the b64.py it seems like it could work by simply adding an additional file extension? Before I went down the rabbit hole I was curious if SVG's presented more nuance than I understand or why you may have not supported it initially

facelessuser:
SVG isn't traditionally base 64 encoded.

jmdye:
Can this also be made to work with reference style image links? I am have created a markdown file continaing reference-style image links that resolve to my inline icons, then I'm using snippets to inject it at the end of a markdown file. I wanted to use b64 so that I didn't have to distribute the icons themselves in the built site, but it appears to not work with reference style image links.

facelessuser:
@jmdye, I'd need you to provide an example of what you mean. I'm not sure I am following.

jmdye:
> @jmdye, I'd need you to provide an example of what you mean. I'm not sure I am following.

sorry. I'm just dumb. Bad syntax on my part and I jumped the gun. it works fine with reference style images. SVG support will be sweet, looking forward to that.

Example fwiw:

**my-doc.md**
```
Lorem ipsum dolor sit amet, ![icon][icon-one] consectetur adipiscing elit.

--8<-- "icons.md"
```

**icons.md**
```
[icon-one]: icons/inline/consectetur.png
[icon-two]: icons/inline/icon2.png
[icon-three]: icons/inline/icon3.png
```

facelessuser:
> sorry. I'm just dumb. Bad syntax on my part and I jumped the gun. it works fine with reference style images. 

Yeah, currently it doesn't care about the syntax, it is more looking for images in the HTML tree after that part of the parsing is done.

> Also, I know it's called b64 for a reason, but data URIs for SVG would probably work just fine as well.

[...]

PR #1619, opened 2022-02-24
b64: Add svg to list of supported extensions

mtdcr:
Tested and works fine.

### DIFF
--- a/pymdownx/b64.py
+++ b/pymdownx/b64.py
@@ -35,7 +35,8 @@ RE_SLASH_WIN_DRIVE = re.compile(r"^/[A-Za-z]{1}:/.*")
 file_types = {
     (".png",): "image/png",
     (".jpg", ".jpeg"): "image/jpeg",
-    (".gif",): "image/gif"
+    (".gif",): "image/gif",
+    (".svg",): "image/svg+xml",
 }
 
 RE_TAG_HTML = re.compile(


facelessuser:
It's not really the ideal way to embed SVG, but I guess it does work. You'd almost be better off embedding the SVG raw as it would be smaller in size than being converted to base64. There are other ways to minimize SVG and embed them.

facelessuser:
We'd probably need tests for this if we were to accept it, but I'm honestly not sure how I feel about embedding SVG with Base64 as it just increases the size even more. Especially when you don't need to base64 SVG.

mtdcr:
Of course size increases with base64, but it does increase for png, too. It's surprising for a user if some formats get inlined and others don't. There's no other plugin that inlines svg img tags currently, AFAICS. I don't have control over the types of images used in my application or how they get embedded. I'd just like to inline them.

There already are tests for png covering the blob-to-base64 aspect, and none for the other supported file types. I don't expect any difference at all. Which aspect would you like to get covered?

mtdcr:
See also https://stackoverflow.com/a/21626701 for funny problems arising when trying to avoid base64-encoding svg in data uris.

facelessuser:
>  but it does increase for png, too.

Yeah, but you can't embed a PNG directly without base64, that's why it's a fine option. For this reason, I considered a general-purpose extension to replace them, a general-purpose image embedding extension: https://github.com/facelessuser/pymdown-extensions/issues/862.

With that said, less desirable doesn't mean undesirable 🤷🏻.

> There already are tests for png covering the blob-to-base64 aspect, and none for the other supported file types. I don't expect any difference at all. Which aspect would you like to get covered?

I guess that's a fair argument 🙂.

facelessuser:
I'll probably let it go in. Even if we do this better in a future extension, there isn't really a reason to not allow it here.

facelessuser:
@gir-bot lgtm

PR #1623, opened 2022-02-25
Fixes extended Pygments language case-sensitive issue

### LINKED ISSUE #1622: Highlight: Improve documentation on extend_pygments_lang
### Description

While using the `extend_pygments_lang` option of the Highlight extension, I faced some minor issues:

1. [This](https://github.com/facelessuser/pymdown-extensions/blob/bbe726b498453ef9e40a9730e5a6e7684614f20a/docs/src/markdown/extensions/highlight.md?plain=1#L77) line in the docs should read "extend_pygments_lang" instead of "extend**ed**_pygments_lang".
2. It wasn't clear to me how to translate the config example to proper YML code for `mkdocs.yml`. Luckily, I found [this test](https://github.com/facelessuser/pymdown-extensions/blob/439c386976345c666189665af2c88fe0b8dfd60a/tests/extensions/superfences/tests.yml#L54) in your sources, that helped a lot. 
3. The `name` value is case-sensitive, which wasn't expected. So if you use, e.g., `PHP-inline` in your code and define `name: php-inline`, syntax highlighting won't work. Normally, the language aliases are case insensitive. (Not sure whether that's a Pygments issue or can be fixed in your extension.)

### Minimal Reproduction

See description.

### Version(s) & System Info

- Operating System: macOS Monterey
- Python Version: 3.9.10
- Package Version: 9.0

facelessuser:
Fixes #1622

### DIFF
--- a/docs/src/dictionary/en-custom.txt
+++ b/docs/src/dictionary/en-custom.txt
@@ -156,6 +156,7 @@ html
 indepth
 injectable
 inline
+inlined
 inlines
 ish
 isn't
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 9.3
+
+- **NEW**: B64: Allow SVG to be encoded and inlined.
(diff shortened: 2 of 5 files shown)


facelessuser:
@gir-bot lgtm

PR #1644, opened 2022-03-24
Use Hatch for build backend

facelessuser:
Supersedes #1640


pyproject.toml
```diff
@@ -1,7 +1,66 @@
 [build-system]
 requires = [
-    "setuptools>=42",
-    "wheel"
+    "hatchling>=0.21.0",
```
ofek, line 3: Latest:

```suggestion
    "hatchling>=0.21.1",
```

ofek (changes requested):

ofek (approved):

facelessuser:
I was holding out on this as my Python 3.10 would not work with this change. Turned out that the brew install on my mac was just busted in a lot of ways (I hadn't switched over to 3.10 for this and many reasons). So I finally installed the official distribution, and everything worked...go figure.

I may be merging this soon. I realize this is a new build backend, and there are probably not a lot of people using it, so that does give me a little pause, but it does seem to work quite nicely.

I am curious though about what specifically Hatch offers that others don't. I hadn't looked into build-backends until this point, but there are packages like https://github.com/pypa/flit maintained by pypa.  What makes Hatch better than others that have a larger community maintaining them?

ofek:
I can briefly tell you why we moved at Datadog!

### setuptools

Disadvantages:

- Support for `pyproject.toml`-based metadata declaration is experimental atm
- No support for editable installation standards ([PEP 660](https://www.python.org/dev/peps/pep-0660/)) so `pip install -e ...` still requires a `setup.py` file
- Based on a [deprecated](https://www.python.org/dev/peps/pep-0632/) standard library module that it now must vendor
- Slow to get things merged since the [code quality issue](https://github.com/pypa/pipx/pull/817#issuecomment-1050255755) is exacerbated by the fact that it is mostly maintained by just [one person](https://github.com/jaraco)

### Flit

[`flit-core`](https://pypi.org/project/flit-core/) is the build backend behind [Flit](https://github.com/pypa/flit).

Disadvantages:

- No mechanism for extending behavior by design, so we could never do things like versioning with Git tags
- Very opinionated about project layouts since it targets beginners e.g. [only supports](https://github.com/pypa/flit/issues/370#issuecomment-973359998) implicit namespace packages, must define certain data like `__version__` inside the root `__init__.py` file, etc.

### Poetry

[`poetry-core`](https://pypi.org/project/poetry-core/) is the build backend behind [Poetry](https://github.com/python-poetry/poetry).

Disadvantages:

- Persistent unwillingness to adopt new standards if they are deemed suboptimal (see comments on [PEP 621](https://discuss.python.org/t/5243/12) and [PEP 665](https://github.com/python-poetry/poetry/issues/4710#issuecomment-973946104))

### Hatchling

- In comparison to `setuptools`, Hatchling has reproducible builds, supports the modern [PEP 660](https://peps.python.org/pep-0660/) editable installation standard, has a more intuitive plugin system, and the codebase is an order of magnitude smaller thus making it more auditable and less prone to bugs. For example, the logic to build a wheel resides in this [one file](https://github.com/ofek/hatch/blob/master/backend/src/hatchling/builders/wheel.py).
- In comparison to `flit-core`, Hatchling is extensible by design. For example, there is already [a plugin](https://github.com/ofek/hatch-vcs) for [setuptools_scm](https://github.com/pypa/setuptools_scm) (which despite the legacy naming is actually now decoupled from `setuptools`) and [a plugin](https://github.com/ofek/hatch-mypyc) for compiling code with [Mypyc](https://github.com/mypyc/mypyc).
- In comparison to `poetry-core`, Hatchling never strays from community standards and is in fact eager to adopt new ones (it's the only backend that currently supports [PEP 639](https://www.python.org/dev/peps/pep-0639/)).

ofek:
I don't have a comprehensive list of users yet, but more projects switched recently too like [tox](https://github.com/tox-dev/tox/blob/rewrite/pyproject.toml#L3)

facelessuser:
Hey thanks for the info. Sounds like hatch is the way to go. Thanks for your help and patience.

facelessuser:
@gir-bot lgtm

PR #1656, opened 2022-04-04
Section Snippets

GDYendell:
This is an addition to include just one section of a file, given by the heading.

I see there are a few feature requests floating around and that it is uncertain what direction this extension will take. I hope this is a simple enough change to be considered and iterated upon.

### DIFF
--- a/pymdownx/snippets.py
+++ b/pymdownx/snippets.py
@@ -27,6 +27,7 @@ from markdown.preprocessors import Preprocessor
 import re
 import codecs
 import os
+from typing import List
 
 
 class SnippetPreprocessor(Preprocessor):
@@ -120,6 +121,10 @@ class SnippetPreprocessor(Preprocessor):
                 # Get spaces and snippet path.  Remove quotes if inline.
                 space = m.group('space').expandtabs(self.tab_length)
                 path = m.group('snippet')[1:-1].strip() if inline else m.group('snippet').strip()
+                if "|" in path:
+                    path, section = path.split("|")
+                else:
+                    section = ""
 
                 if not inline:
                     # Block path handling
@@ -142,9 +147,14 @@ class SnippetPreprocessor(Preprocessor):
                         self.seen.add(file_name)
                     try:
                         with codecs.open(snippet, 'r', encoding=self.encoding) as f:
-                            new_lines.extend(
-                                [space + l2 for l2 in self.parse_snippets([l.rstrip('\r\n') for l in f], snippet)]
-                            )
+                            content = f.readlines()
+
+                        if section:
+                            content = self.extract_section(content, section)
+
+                        new_lines.extend(
+                            [space + l2 for l2 in self.parse_snippets([l.rstrip('\r\n') for l in content], snippet)]
+                        )
                     except Exception:  # pragma: no cover
                         pass
                     if file_name:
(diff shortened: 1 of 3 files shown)


facelessuser:
Thanks for your interest in creating a pull request!

Unfortunately, I do have some concerns. I imagine this will be quite fragile. How does it know when a section is in a code block or out of a code block? Can it tell the difference when under a list vs in an indented code block?

There are likely a number of ways this could break as it doesn't have the full Markdown context. It would be better if the syntax was very unique and could be processed without Markdown context, but since this is using the Markdown syntax itself, it becomes problematic.

For this reason, you almost have to build up a template language as you add features like this. At that point, it would make more sense to just use an existing templating library than to build up your own. Something like Jinja2 could do this, and you could customize the opening and closing indicators to work optimally in Markdown.

Anyways, this is kind of why I've been sitting on this one. Once I add something like this, I have to maintain it. I'm afraid this would break in a number of ways that would be somewhat complex to fix.

GDYendell:
I understand your concerns - even such a simple example throws up some edge cases I didn't think of initially!

What if it used a commonmark compliant markdown parser and formatter, such as [markdown_it](https://github.com/executablebooks/markdown-it-py) and [mdformat](https://github.com/executablebooks/mdformat) (or other as you prefer), instead of parsing the raw text? So, `extract_section` would become (this works locally for my specific case):

```
    def extract_section(self, lines: List[str], section: str) -> List[str]:
        """Return a subset of lines for the given section

        If section is not found, all lines are returned.
        """
        HEADING = "heading_open"
        start = 0
        end = -1

        text = "\n".join(lines)

        md = MarkdownIt()
        tokens = md.parse(text)

        start_level = -1
        for idx, token in enumerate(tokens):
            if token.type == HEADING and section in tokens[idx+1].content:
                start = idx + 3  # Skip heading_open, heading content, header_close
                start_level = len(token.markup)
                break

        for idx, token in enumerate(tokens[start:]):
            if token.type == HEADING and len(token.markup) <= start_level:
                end = start + idx  # Include everything up and not including to this heading
                break

        content = MDRenderer().render(tokens[start:end], md.options, {}).split("\n\n")

        return content
```

Do you think this is more viable, or is there just too much that can go wrong to do this in a reasonable way?

I should also explain at this point what I am trying to do is include a function definition from doxybook generated APIs into prose describing how to implement it, so don't have control of the formatting of the snippets I am trying to include.

I think I can pull this into something custom that works just for what I need, but it would be great if there was a standard way to do it. Perhaps that place is not in this module.

Happy to close this if you are not convinced!

facelessuser:
This is kind of what I'm talking about. Snippets was originally designed so that I could do simple things like keep abbreviation references and link references in one place and include them wherever I needed them. It was also useful for re-used content, or maybe pre-generated content you just wanted to link to. The idea behind it was straightforward and simple, and its behavior was also fairly straightforward and simple.

As time has gone on, people want to do even more with it, and while I understand that desire, I question if Snippets is really the best tool for these more advanced jobs. I feel like people really want something more akin to a Jinja2 plugin, though even that would not have awareness of the Markdown header context that you wish to have.

Personally, I would recommend keeping your sections in separate files and include the file containing the section you desire.

The requirement, in a pre-processing step, to have complete knowledge of all the context of the file, and how they relate to all the included plugins, is just beyond Snippet's ability. Sure we could hack it context-awareness but there would be some costs.

I'm pretty sure you can guess I'm probably not keen on including another Markdown parser in a plugin for yet another Markdown parser so we can have awareness of headers. Why not just instantiate another instance of the original Markdown parser with the same plugins so it has full awareness of any and all custom syntax? This is surely possible, and I've done similar things for some [personal reasons](https://github.com/facelessuser/pymdown-extensions/blob/main/tools/pymdownx_md_render.py#L62). Granted, this would parse the snippet's content in a vacuum, and it would not share the context with the including parent, but it could work as long as you were willing to give up on the idea of the snippet sharing context with the including parent.

I'm not really sure I want to add all of this complexity, spawning parsers all the way down as each snippet includes another snippet, just so Snippets can be aware of headers. This adds quite a bit of complexity and changes some of the assumptions of an imported snippet.

PR #1670, opened 2022-04-26
Fix Pygments version requirement

Menci:
In the latest [release](https://github.com/facelessuser/pymdown-extensions/releases/tag/9.4) this package broke the compatibility with `Pygments < 2.12`. I think we should also change the version range in `requirements`.


facelessuser:
@gir-bot lgtm

PR #1702, opened 2022-05-17
feat: block language class

johanneswuerbach:
Expose the language name used in code block as css class.

Our use-case is doing some client-side processing of mermaid js code blocks, where it would be useful to have access to this information in the rendered output.

### DIFF
--- a/docs/src/markdown/extensions/highlight.md
+++ b/docs/src/markdown/extensions/highlight.md
@@ -79,9 +79,6 @@ wish to apply.
         ]
         ```
 
-!!! new "New 9.2"
-    `pygments_lang_class` added in 9.2.
-
 ## Line Number Styles
 
 Pygments has two available styles when outputting source code with line numbers enabled: `table` and `inline`. `table`
@@ -110,7 +107,7 @@ apply when Pygments is disabled. Many of these options are demonstrated in the [
 Option                    | Type   | Default               | Description
 ------------------------- | ------ | ----------------------| -----------
 `css_class`               | string | `#!py3 'highlight`    | Default class to apply to the wrapper element on code blocks. Other extensions can override this.
-`guess_lang`              | bool   | `#!py3 False`         | Guess what syntax language should be used if no language is specified. 
+`guess_lang`              | bool   | `#!py3 False`         | Guess what syntax language should be used if no language is specified.
 `pygments_style`          | string | `#!py3 'default'`     | Set the Pygments' style to use.  This really only has an effect when used with `noclasses`.
 `noclasses`               | bool   | `#!py3 False`         | This will cause the styles to directly be written to the tag's style attribute instead of requiring a stylesheet.
 `use_pygments`            | bool   | `#!py3 True`          | Controls whether Pygments (if available) is used to style the code, or if the code will just be escaped and prepped for a JavaScript syntax highlighter.
@@ -127,6 +124,7 @@ Option                    | Type   | Default               | Description
 `anchor_linenums`         | bool   | `#!py3 False`         | Enables the Pygments option of a similar name. If set to `#!py True`, will wrap line numbers in `#!html <a>` tags. Used in combination with `linenums` and `line_anchors`. If `line_anchors` is not configured, `__codelineno` will be assumed as the ID prefix.
 `line_anchors`            | bool   | `#!py3 False`         | Controls the Pygments option of a similar name. If set to a nonempty string, e.g. `foo`, the formatter will insert an anchor tag with an id (and name) of `foo-<code_block_number>-<line_number>`.
 `pygments_lang_class`     | bool   | `#!py3 False`         | If set to True, the language name used will be included as a class attached to the element with the associated `language_prefix`.
+`block_lang_class`        | bool   | `#!py3 False`         | If set to True, the block language name will be included as a class attached to the element with the associated `language_prefix`.
 
 !!! new "New 7.1"
     `linenums_class` was added in `7.1`.
@@ -138,3 +136,9 @@ Option                    | Type   | Default               | Description
 
 !!! new "New 9.0"
     `auto_tile`, `auto_title_map`, `line_spans`, `anchor_linenums`, and `line_anchors` were all added in `9.0`.
+
+!!! new "New 9.2"
+    `pygments_lang_class` added in 9.2.
+
+!!! new "New 9.5"
+    `block_lang_class` added in 9.5.
--- a/pymdownx/highlight.py
+++ b/pymdownx/highlight.py
@@ -124,6 +124,10 @@ DEFAULT_CONFIG = {
         False,
         'If set to True, the language name used will be included as a class attached to the element. - Defaults: False'
     ],
+    'block_lang_class': [
+        False,
+        'If set to True, the block language name used will be included as a class attached to the element. - Defaults: False'
+    ],
     '_enabled': [
         True,
         'Used internally to communicate if extension has been explicitly enabled - Default: False'
@@ -230,7 +234,7 @@ class Highlight(object):
         noclasses=False, extend_pygments_lang=None, linenums=None, linenums_special=-1,
         linenums_style='table', linenums_class='linenums', language_prefix='language-',
         code_attr_on_pre=False, auto_title=False, auto_title_map=None, line_spans='',
-        anchor_linenums=False, line_anchors='', pygments_lang_class=False
+        anchor_linenums=False, line_anchors='', pygments_lang_class=False, block_lang_class=False,
     ):
         """Initialize."""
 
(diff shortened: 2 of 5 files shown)


facelessuser:
I'm not understanding what this feature is. Why are you not using `pygments_lang_class`? You've seemed to copy everything that `pygments_lang_class` is doing.

johanneswuerbach:
> You've seemed to copy everything that `pygments_lang_class` is doing.

An alternative could be to render both (the lexer and the original language) as class names if `pygments_lang_class=True`, would that be something acceptable?

facelessuser:
Ah, this sounds like you need to use a custom fence instead. This usage is a very niche, overly specific use case, but you can do this with your own custom fence.

facelessuser:
The reason Pygments doesn't return `language-mermaid` is that it doesn't recognize it and ends up defaulting to something else, in this case, plain text.

johanneswuerbach:
The problem with custom fences is that those would require additional docs configuration, which is something I would like to avoid. 

This issue is part of a larger project to build a mermaid rendered in backstage https://github.com/backstage/backstage/issues/4123#issuecomment-1111530166 and should ideally work fully client side. Right now I rely on a regex to identify mermaid code, but it would be much easier if the user specified language is just exposed here.

facelessuser:
So you want to do something custom, but don't want a custom configuration 🙃.

In all seriousness though, the approach I took was that Highlight will highlight code with Pygments or format the code blocks for JS highlighters. With Pygments though, we expose the language name specified if, and only if, it is recognized. If it is not recognized, we either use language guessing (if enabled) or default to plain text as a last resort. Anything else is outside the scope of Highlight. In your case, `mermaid` is not recognized and thrown away.

SuperFences allows you to hijack the fence format and do custom stuff. Python Markdown was kind of a pain to get nested fences working under lists and other constructs. It sucked enough that I didn't want to do it again, so I created a hook so that anything else that I wanted to do in a container like a fence could just be done with a custom fence. Yes, it is built on top of Highlight, and that is its default configuration, but it can be used for whatever else you want. What you are asking for falls under this scope.

I'm not sure I'm interested in changing the scope of Highlight to propagate invalid language names so people can do custom stuff. I've already added hooks in SuperFence to accommodate people's more advanced needs, and this is my recommendation. This is the path of least resistance for me. It doesn't make sense for me to maintain another custom path when we have a path that can already accommodate a user. Yes, it requires the user to manage their custom stuff a bit more, but that was kind of the intention, less stuff for me to maintain.

I eat my own dogfood as well, and this how I maintain my own custom requirements, for instance, live editable code blocks: https://facelessuser.github.io/coloraide/#what-is-coloraide, custom fences that create a sandbox that executes a Markdown block with different settings: https://github.com/facelessuser/pymdown-extensions/blob/main/docs/src/markdown/extensions/critic.md#usage ([src](https://github.com/facelessuser/pymdown-extensions/blob/main/tools/pymdownx_md_render.py)).

johanneswuerbach:
> So you want to do something custom, but don't want a custom configuration 🙃.

Yes and no. Explaining backstage users how to add an addon is fairly easy, but mkdocs configuration is mostly abstracted away. I totally agree that this isn't your problem, I just wondered if this plugin could expose the already parsed language information in some way (class, data attribute, etc.) to the rendered output. If the answer is no, I can happily close this PR and will continue to rely on the regex.

facelessuser:
I hope I didn't come off in the wrong way. I was kind of being a bit light-hearted, but I know that sometimes things in text can come off a little odd.

If we exposed the parsed language in this instance, I feel it would be misleading in regards to the intent of the feature. I can understand the desire to exploit it in your specific case, but generally, it doesn't make sense in the context in which the extension works. In addition, I just feel like it doesn't serve the wider audience either, only backstage in this case.

johanneswuerbach:
👍 I‘ll close this PR in this case :-)

webark:
@facelessuser Just came across this cause I had a similar use case. What if instead of adding the same `language-<lang>` class, we added an `unrecognized-language-<lang>`, or just `unrecognized-language` for the class, but added a data attribute with the value the individual had specified?

🤔 I guess you would potentially need to do a level of sanitation there, cause you would be placing a user generated string in the HTML, but that might need to be done regardless.

Anyway, let me know if this variation on this PR "cuts the mustard" enough for ya. 🌭

facelessuser:
@webark I'm still having extreme difficulty understanding what this is trying to solve. We already have custom fences, so if you need to have a specific fence recognized, it is super easy to create one and even have it output as a code block. The opening post claimed they wanted to do this for mermaid, but this can already be done. I do this for mermaid diagrams myself: https://github.com/facelessuser/pymdown-extensions/blob/main/mkdocs.yml#L94. I use `diagram`, but one could use `language-mermaid` or anything else.

I'm not sure what `unrecognized-language-<lang>` is going to solve for the majority of people. The majority of people just want code to get highlighted, and if it can't be, it `text` is fine. If they have specific needs for a specific block, this can easily be done by just creating a special, custom fence for it, not code needed if you just want it wrapped as a code block or a `div` as we already provide custom fences for that. How many non-language blocks do you have to handle that will not be recognized? Can you explain how this would help you?

I will state that it may be a hard sell for me to add a new feature that only 0.5% of users would use, especially when I think most issues can already be mitigated, but I do want to understand exactly what is the current struggle, and how do you envision this proposed feature to help?

webark:
@facelessuser Thanks for being responsive so far.

And thanks for the info about super fences. That ideally would be the appropriate approach I believe. It might take some lobbying to get a default generic fence added to the general mkdocs renderer, but it is a more appropriate approach, rather than have it be translated into a highlighted code block, and then grabbing it out of that. I will look into that that and just to "put a bow on it" report back my experience. 🎀

The one more point that I will say, and why I stumbled down this approach originally, is that as a user, I was surprised that if I were to add "mermaid" or "custom-mustache-dsl" or what have you, that this value would get parsed out, removed completely, and left with just "language-text", but if there happened to be a matching one (which i'm unsure where that comprehensive default list is) it would get added after the "language-". And even though as I understand it that is intentional and desired on your part, it is unexpected and alarming that this content would just be removed from the output completely.

Anyway. Thanks for the time and the point towards superfences. I'll try it out and see how it goes.

facelessuser:
> The one more point that I will say, and why I stumbled down this approach originally, is that as a user, I was surprised that if I were to add "mermaid" or "custom-mustache-dsl" or what have you, that this value would get parsed out, removed completely, and left with just "language-text", but if there happened to be a matching one (which i'm unsure where that comprehensive default list is) it would get added after the "language-". And even though as I understand it that is intentional and desired on your part, it is unexpected and alarming that this content would just be removed from the output completely.

Okay, let me explain things to try and make clear why we currently take the approach we take.

The whole reason for inventing a standardized approach with code blocks with special classes such as `language-<lang>` in HTML is for JS syntax highlighters so they can find and highlight the languages as required.

Pygments is not a JS highlighter, instead, it is done on the backend. The language you apply to a fence is used to instruct Pygments on what highlighter to use. Pygments, not us, generates HTML with no language class because it is no longer needed. Pygments, if enabled, will even guess what language you need if you don't apply a language to the fence. If guessing is not enabled, at the very least, an unknown or missing language is treated as plain text.

Because Pygments determines that the language class is not needed, we added the `pygments_lang_class` class to expose the class and provide some information back to the user as to what the highlight code block is. Keep in mind that Pygments may guess the language, so returning what Pygments decided is a conscious choice to create a generic, meaningful return. In this case, the language class takes on a new meaning and represents what Pygments has decided as the language instead of trying to inform a JS highlighter what language of highlighter is desired.

Now, the complaint is that the Highlight plugin should just carry over the `mermaid` language (or whatever unknown language) and then everything will work as expected, but this is short-sighted and I'll explain why. You may rely on this `mermaid` class (or insert whatever special language class you've decided on) right now to do some post-processing and provide your own special language translation. But what if Pygments in the future adds a highlighter for Mermaid diagram syntax? What will you do then? You'll have all your diagrams silently broken one day because Pygments will translate them to highlighted code blocks that can no longer be able to process them in your very special way. If we don't add a special "unknown" attribute, you'll just get back an incompatible, highlighted code block. If we do add the "unknown" attribute, it'll just disappear one day and break all your blocks.

By not explicitly handling your special blocks, you create a fragile system that may break in the future. This is my personal feeling, and it of course shapes the way I approach this plugin. If you want to handle something special, you need to explicitly handle it specially. This way there are no surprises down the line.

The average user is not going to have an unknown number of unrecognized languages they need to process special, on the contrary, they will most certainly have a very finite, well-known amount of special blocks. They should define those special blocks and handle them as they desire. The user may not even want to handle them as code blocks. For instance, Mermaid specifically expects the contents to be wrapped in `<div>` by default. There may be other formats for special blocks where a JS translator expects the block to not be a code block. SuperFences custom fences were created for this purpose and it provides the necessary hook to allow you to specially handle and pre-process them however you need them to be without fear of Pygments one day utilizing your specially handled language tag.

webark:
I wouldn't say complaint, more surprised. Kind of like if I had ordered just "a salad" and they brought out a fruit salad, it could be equally if not more tasty, but just unexpected. And if it did start to process, the unknown class/flag would drop off and it would just skip it, or there would be no code block and it would skip it as well.

But i do agree that custom fences is the better approach, and should allow enough communication to the browser for us to add whatever dynamic renderer we need to.

Thanks again! 😊

PR #1725, opened 2022-06-04
Allow manually specifying a tab

### LINKED ISSUE #1724: Ability to focus on a given tab in a tab set.
### Description

Currently, there is no way to focus on a given tab (among tabs). The first tab always gets the focus. It would be nice if there was a way to put focus on a specific tab.

Relevant `mkdocs-material` discussion https://github.com/squidfunk/mkdocs-material/discussions/3982

### Benefits

One use case I have is described here https://github.com/squidfunk/mkdocs-material/discussions/3982#discussioncomment-2881853

### Solution Idea

One possible solution using `===+` is described here https://github.com/squidfunk/mkdocs-material/discussions/3982#discussioncomment-2881832

facelessuser:
Resolves #1724

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 9.7
+
+- **NEW**: Tabbed: Add new syntax to allow forcing a specific tab to be select by default.
+
 ## 9.6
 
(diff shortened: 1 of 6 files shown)


facelessuser:
Need to figure out what to do in docs as we wanted to use `content.tabs.link` in our docs, but we can't demonstrate tab selection and demonstrate tab linking at the same time due to the current implementation.

facelessuser:
@gir-bot lgtm

PR #1728, opened 2022-06-07
Allow only guessing language for inline/block code

### LINKED ISSUE #1696: Allow overriding language guessing in InlineHilite
### Description

The language guessing in InlineHilite is controlled by `highlight.guess_lang`, and cannot be turned off as long as that option is true. The current `style_plain_text` option allows one to override `guess_lang` to true, but not the other way around.

### Benefits

It should be a legitimate use case where language guessing is desired for code blocks but not for inline code.

### Solution Idea

`style_plain_text` should be deprecated in favor of another option (perhaps also named `guess_lang`) which would exclusively control the language guessing behaviour for InlineHilite.

SuibianP:
There was no way to specify the type of code block to be guessed language, for example, to only guess language for code blocks but not inline code, the length of which is usually insufficient for satisfactory accuracy.

Allow 'inline' and 'block' to be specified for `guess_lang` in addition to boolean values.

Fixes #1696.

### DIFF
--- a/pymdownx/highlight.py
+++ b/pymdownx/highlight.py
@@ -52,7 +52,7 @@ DEFAULT_CONFIG = {
         'Default: True'
     ],
     'guess_lang': [
-        False,
+        0,
         "Automatic language detection - Default: False"
     ],
     'css_class': [
@@ -484,7 +484,7 @@ class HighlightTreeprocessor(Treeprocessor):
 
                 self.ext.pygments_code_block += 1
                 code = Highlight(
-                    guess_lang=self.config['guess_lang'],
+                    guess_lang=(self.config['guess_lang'] == True) or (self.config['guess_lang'] == 'block'),
                     pygments_style=self.config['pygments_style'],
                     use_pygments=self.config['use_pygments'],
                     noclasses=self.config['noclasses'],
--- a/pymdownx/highlight.py
+++ b/pymdownx/highlight.py
@@ -291,7 +291,7 @@ class Highlight(object):
             lexer = None
 
         if lexer is None:
-            if self.guess_lang or (self.guess_lang == 'inline' if inline else self.guess_lang == 'block'):
+            if (self.guess_lang is True) or (self.guess_lang == 'inline' if inline else self.guess_lang == 'block'):
                 try:
                     lexer = guess_lexer(src)
                     name = lexer.aliases[0]
--- a/pymdownx/highlight.py
+++ b/pymdownx/highlight.py
@@ -274,7 +274,7 @@ class Highlight(object):
 
         return self.extend_pygments_lang.get(language.lower(), (language, {}))
 
-    def get_lexer(self, src, language):
+    def get_lexer(self, src, language, inline):
         """Get the Pygments lexer."""
 
         name = language
@@ -291,7 +291,7 @@ class Highlight(object):
             lexer = None
 
         if lexer is None:
-            if self.guess_lang:
+            if self.guess_lang or (self.guess_lang == 'inline' if inline else self.guess_lang == 'block'):
                 try:
                     lexer = guess_lexer(src)
                     name = lexer.aliases[0]
@@ -332,7 +332,7 @@ class Highlight(object):
                 raise RuntimeError('Pymdownx Highlight requires at least Pygments 2.12+ if enabling Pygments')
 
             # Setup language lexer.
-            lexer, lang_name = self.get_lexer(src, language)
+            lexer, lang_name = self.get_lexer(src, language, inline)
             if self.pygments_lang_class:
                 class_names.insert(0, self.language_prefix + lang_name)
             linenums = self.linenums_style if linenums_enabled else False
@@ -484,7 +484,7 @@ class HighlightTreeprocessor(Treeprocessor):
 
                 self.ext.pygments_code_block += 1
                 code = Highlight(
-                    guess_lang=(self.config['guess_lang'] == True) or (self.config['guess_lang'] == 'block'),
+                    guess_lang=self.config['guess_lang'],
                     pygments_style=self.config['pygments_style'],
                     use_pygments=self.config['use_pygments'],
                     noclasses=self.config['noclasses'],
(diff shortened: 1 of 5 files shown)


pymdownx/highlight.py
```diff
@@ -52,7 +52,7 @@
         'Default: True'
     ],
     'guess_lang': [
-        False,
+        0,
```
facelessuser, line 55: I think what would make sense is to set this to `None`. If `None`, we will assume the default of `False`. Otherwise, we will accept booleans or `block` or `inline`. No other strings should be accepted.

facelessuser, line 55: I think what would make sense is to set this to `None`. If `None`, we will assume the default of `False`. Otherwise, we will accept booleans or `block` or `inline`. No other strings should be accepted.

SuibianP: > No other strings should be accepted.

This makes sense, will reject all other strings not parsed as boolean and not `'inline'` or `'block'`.

> I think what would make sense is to set this to `None`. If `None`, we will assume the default of `False`. Otherwise, we will accept booleans or `block` or `inline`.

I am not sure what benefits would `None` offer here. I think `0` here is functionally equivalent to `False` here except that non-boolean `str` inputs will be allowed? I chose `0` because it is also used in this way by the official `toc` extension [here](https://github.com/Python-Markdown/markdown/blob/a767b2daaad78ba32d45a4f1dabb7c5e218f030a/markdown/extensions/toc.py#L341-L343).

facelessuser: > I am not sure what benefits would None offer here. I think 0 here is functionally equivalent to False here except that non-boolean str inputs will be allowed? I chose 0 because it is also used in this way by the official toc extension [here](https://github.com/Python-Markdown/markdown/blob/a767b2daaad78ba32d45a4f1dabb7c5e218f030a/markdown/extensions/toc.py#L341-L343).

I could also show examples in Python Markdown that use None, consider [`codehilite`](https://github.com/Python-Markdown/markdown/blob/a767b2daaad78ba32d45a4f1dabb7c5e218f030a/markdown/extensions/codehilite.py#L280).  Probably more important, we have an example in the very file you are editing [here](https://github.com/facelessuser/pymdown-extensions/blob/main/pymdownx/highlight.py#L72).

Providing a specific type like boolean will cause Python Markdown to type check the options. Why Pyhton Markdown does not type check integers is anyone's guess, but usually, when providing an optional parameter, `None` is [used](https://docs.python.org/3/library/typing.html#typing.Optional), so this makes more sense to me and is my personal preference for this project.

SuibianP: It seems that providing `None` only allows a tri-state (`True`/`False`/`None`), but not any other types such as `str`. The relevant snippet is [here](https://github.com/Python-Markdown/markdown/blob/a767b2daaad78ba32d45a4f1dabb7c5e218f030a/markdown/extensions/__init__.py#L60-L61).

The local example of `linenums` is fine because it does not go beyond the three choices (try giving it a string and it will fail in the superclass constructor), and the `codehilite` example gets away by not using the superclass constructor at all.

We can take the latter approach, but it might not be really worth it. Given that `inline` is hardly of any practical use, can we <ins>ab</ins>use the third state of `None` as `block` instead?

facelessuser: > It seems that providing None only allows a tri-state (True/False/None), but not any other types such as str. The relevant snippet is [here](https://github.com/Python-Markdown/markdown/blob/a767b2daaad78ba32d45a4f1dabb7c5e218f030a/markdown/extensions/__init__.py#L60-L61).

Ugh, that is awful. Oh well, I guess we leave it as zero.

facelessuser: > We can take the latter approach, but it might not be really worth it. Given that inline is hardly of any practical use, can we abuse the third state of None as block instead?

Well, that is a good question. It isn't quite intuitive though. I need to think about this a bit. One option is just to ignore guessing in inline, but I don't know if people actually use guessing for inline...We can probably just use zero as the default, even though I don't like it. Our hands are kind of tied due to the way Python Markdown approaches things like `None`.

pymdownx/inlinehilite.py
```diff
@@ -126,7 +126,7 @@ def get_settings(self):
             self.css_class = css_class if css_class else config['css_class']
 
             self.extend_pygments_lang = config.get('extend_pygments_lang', None)
-            self.guess_lang = config['guess_lang']
+            self.guess_lang = (config['guess_lang'] == True) or (config['guess_lang'] == 'inline')
```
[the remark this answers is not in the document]
facelessuser: The Highlight class itself should be the only thing that processes the input. It is the only thing that should be aware.  We instantiate the Highlight class in InlineHilite only when it hasn't already been included as it is dependent on it, but the logic should be contained in Highlight.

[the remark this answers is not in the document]
SuibianP: Fixed.

On a side note, I think InlineHilite instantiates its own `Highlight` regardlessly.

https://github.com/facelessuser/pymdown-extensions/blob/ea60333049783280dd90b304a670da12ff853a67/pymdownx/inlinehilite.py#L139-L150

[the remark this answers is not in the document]
facelessuser: It does. It was modeled after SuperFences which allows a few inline overrides via fence code options, such as line numbers on a per block basis, but generally, handling of that logic is performed in the Highlight class.

There was some uncertainty as to whether we would need a per inline override. I've worked to avoid it.

tests/test_extensions/test_highlight.py
```diff
@@ -31,6 +31,35 @@ def test_guess(self):
         )
 
 
+class TestHighlightGuessBlock(util.MdCase):
+    """Test that highlighting works with guessing for block."""
+
+    extension = ['pymdownx.highlight', 'pymdownx.superfences']
+    extension_configs = {
+        'pymdownx.highlight': {
+            'guess_lang': "block"
```
facelessuser, line 40: We are testing things like `block` or `inline`, but what we really need are tests that also test that when `block` is `True`, inline code does not use guessing and vice versa. That is the only way to ensure the feature is working as expected.

facelessuser:
@gir-bot lgtm

PR #1734, opened 2022-06-13
Suppress expected deprecation warnings from test

SuibianP:
There are tests specifically targeting legacy/deprecated features, where `DeprecationWarning`'s are expected and should not be displayed.

Suppress these with pytest markers to reduce noise.

https://docs.pytest.org/en/6.2.x/warnings.html#pytest-mark-filterwarnings

### DIFF
--- a/tests/test_extensions/test_emoji.py
+++ b/tests/test_extensions/test_emoji.py
@@ -3,6 +3,7 @@ from .. import util
 import pymdownx.emoji as emoji
 import warnings
 import markdown
+import pytest
 
 
 def _old_style_index():
@@ -23,6 +24,7 @@ def _new_style_index(options, md):
     return index
 
 
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
 class TestEmojiOldIndex(util.MdCase):
     """Test old style index."""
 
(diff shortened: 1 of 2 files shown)


facelessuser:
Hmm, I don't think it's been long enough to remove the deprecated slugify code, so maybe we can ignore that, but maybe the emoji deprecation can be removed. I need to check when it was added.

facelessuser:
I'm not actually ready to remove the emoji deprecation either...I guess we can consider ignoring them.

SuibianP:
With that, I am closing this pull request now.

On a side note, maybe it is possible to follow the deprecation cycle of Python and remove obsolete features when the latest Python version removes the feature in question? This way the tracking would be simpler (removing features when tests on the latest Python errors out).

facelessuser:
> On a side note, maybe it is possible to follow the deprecation cycle of Python and remove obsolete features when the latest Python version removes the feature in question? This way the tracking would be simpler (removing features when tests on the latest Python errors out).

Not quite what you mean. The plan is to have it "deprecated" until I remove it. Neither of these depreciations is tied to a Python feature.

PR #1740, opened 2022-06-24
fix: use get_pymdownx_highlighter consistently

maartenbreddels:
Awesome library, nice work.

I wanted to have custom behavior where markdown code gets executed and add a little widget as output under the code. This can be easily implemented by subclassing the `Highlight` class and overriding `def highlight(...)`. 

However, there was one place where the Highlight class was directly used, instead of via `get_pymdownx_highlighter` (which can also be overridden).

Let me know if you think this is a good usage of your library, and if you think this API will be stable in the future?

FYI, this is how I implement this:

```python
class Highlight(pymdownx.highlight.Highlight):
    def highlight(self, src, language, *args, **kwargs):
        run_src_with_solara = False
        if language == "solara":
            run_src_with_solara = True
            language = "python"
        html = super().highlight(src, language, *args, **kwargs)
        if run_src_with_solara:
            html_widget = run_solara(src)
            return html + html_widget
        else:
            return html


class HighlightExtension(pymdownx.highlight.HighlightExtension):
    def get_pymdownx_highlighter(self):
        return Highlight

```

Regards,

Maarten

### DIFF
--- a/pymdownx/highlight.py
+++ b/pymdownx/highlight.py
@@ -22,6 +22,7 @@ All changes Copyright 2008-2014 The Python Markdown Project
 License: [BSD](http://www.opensource.org/licenses/bsd-license.php)
 """
 import re
+from typing import Type
 from markdown import Extension
 from markdown.treeprocessors import Treeprocessor
 import xml.etree.ElementTree as etree
@@ -483,7 +484,8 @@ class HighlightTreeprocessor(Treeprocessor):
             if len(block) == 1 and block[0].tag == 'code':
 
                 self.ext.pygments_code_block += 1
-                code = Highlight(
+                highlight_class : Type[Highlight] = self.ext.get_pymdownx_highlighter()
+                code = highlight_class(
                     guess_lang=self.config['guess_lang'],
                     pygments_style=self.config['pygments_style'],
                     use_pygments=self.config['use_pygments'],


facelessuser:
A couple of things related to this. While this shouldn't hurt anything, there are a couple of reasons I am hesitant to allow this:

1. This kind of behavior is not documented, and I cannot guarantee in some future revision that this won't be broken. The whole purpose of providing `get_pymdownx_highlight_settings` and `get_pymdownx_highlighter` was to make it easy to expose all relevant settings and get access to the highlighter object in `SuperFences` and `InlineHilite`.  As all of these are controlled within the package that is `pymdown-extensions`, I have full control to ensure if anything changes that I do not break external usage. I have no intention of promising that the approach will not change in some future version as this is an implementation detail, not an exposed API. In short, "Here be dragons".
2. While I have implemented type annotations in many other projects that I support, I have not yet approached them in pymdown-extensions. Without implementing types to a **more** significant degree (not necessarily the entire project) and enabling some kind of verification/validation, I find no useful reason to provide annotations on a single variable within a single function.

facelessuser:
I think what really needs to happen is that custom fences should be used. Here is a discussion on how you could wrap the existing handler with your own: https://github.com/facelessuser/pymdown-extensions/discussions/1739. You can even override the default  behavior by using the name `*` instead of adding specific, new language names.

maartenbreddels:
Perfect, well, at least communicated what I wanted :)

facelessuser:
I'm more than happy to help answer any questions to help you get where you want using the approved documented approach, but providing non-documented, non-tested approaches that circumvent the intended usage can cause problems for me down the road, especially if I end up accidentally breaking it.

Hopefully, I've at least articulated well the reasons behind my hesitancy. I believe there are at least some valid reasons to not **encourage** undocumented behavior, especially since there are documented ways to get the intended results.

maartenbreddels:
Sorry, i think I didn't express myself well. I could implement what I wanted with the link you provided me, and it's a much simpler solution.

Thanks a lot, my problem is solved!

I don't think my PR has any use now, because it doesn't add anything.

facelessuser:
I'm glad you figured it out. It is for these very cases I added these hooks. I eat my own dog food as well and have implemented some very complex solutions with it. See live code execution with color previews: https://facelessuser.github.io/coloraide/. That is why I encourage such approaches.

maartenbreddels:
I didn't make the connection between that system, and the highlighter, is
the highlighter a 'fallback' if no other 'fence handler' is found?

(from mobile phone)

Op za 25 jun. 2022 00:29 schreef Isaac Muse ***@***.***>:

PR #1777, opened 2022-08-06
Feature/directives

facelessuser:
Completely experimental. The idea is to (eventually) replace Admonitions, Details, and Tabs with a new directive format.
We won't be killing off replaced legacy extensions, not for some time.

Pros:
- No indentation is needed.
- Won't mess up your syntax highlighter in your code editor due to said indentation.
- No more inventing new syntax for various new block features.
- We can reuse this syntax for figure captures, embedding videos, etc. Opens the door for cool new features with minimal effort.
- Any existing plugins converted to this format should have outputs that are compatible.
- Would resolve features where people want to add IDs to tabs, Details, etc. We could expose these in the config for the given directive.

Cons: 
- New syntax to learn, but we won't be killing off legacy extensions right away.

What are we waiting on?

- [x] Needs feedback
- [x] Settle on what we call this extension (currently directives, but we could call them anything)
- [x] Settle on the final format/syntax.
- [x] Settle on the feature set/capabilities and API we provide.
- [x] Provide tests and ensure that there aren't any big corner cases we are missing.
- [x] Needs documentation


facelessuser:
I'll duplicate this here as well in order to get feedback:

Just an overall description of how the current prototype is laid out. I think I'm ready to discuss the syntax.

```
:::{directive-name}: arguments
---
option1: value
option2: value
---

content
:::
```

1. Fencing of blocks requires 3 or more `:`. If nesting blocks, the outer block should have a greater length of `:` than the child blocks. Currently, if a closing fence is encountered that is greater than or equal to the starting fence, it is sufficient to close the current fence.
2. `directive-name` is currently the name of whatever directive you are using. It is case insensitive.
3. `arguments` can be one or many arguments, and the given directive would specify required delimiters if required.
4.  The optional frontmatter containing options is YAML-ish bock that comes immediately after the header, or better put, is part of the header when specified. We currently went with YAML-ish to avoid pulling in PyYAML as a requirement. I'm kind of going with simpler is better right now.

    Due to limitations of the Python Markdown parser, it must be a "tight" config (no empty new lines). This also means there should be no newline after the initial header line.

    The key value pairs are similar to the `meta` extension (though I haven't yet made them multi-line, was considering this).

    The directive would impose any specific rules that these options require: delimiter split list, etc.
5. `content`: any valid content can be used, even nested directives. There does not have to be a separation between the header and the content, but can be if desired.

## Examples

Note with no content

```
:::{warning} All we need is a title!
:::
```

```
:::{warning} This is important!
Read this notice.
:::
```

```
:::{warning}
Titles are optional with some directives!
:::
```

```
:::{figure} some/image.png
---
width: 300
---

I'm a figure caption, and you can specify optional settings for your directives as well!
:::
```

```
::::{tip} Here's a secret

:::{note}
You can nest directives as well
:::

::::
```

facelessuser:
We could add the ability for these blocks to insert raw text. That is without an HTML wrapper. Why would we want to do this?

Well, IIRC, Python Markdown does not allow HTML except at the root indentation level. So under lists is impossible.

This should be pretty straightforward. If we mark a directive as being a text-only directive, we could simply store the content in the Python Markdown's stash and attach the placeholder to the end of the parent `text` or its last child's `tail`.

We could even add an event that allows you to process the text, such as formatting it for HTML, or just HTML escape it.

It's probably the only other case I can think of that people _might_ want to do.

facelessuser:
The current syntax is what is up for debate. Do we keep the Myst like approach? Do we do something different? Under the hood, I think the functionality is all there.

facelessuser:
Anyone checking this out should potentially voice their opinion syntax over at https://github.com/Python-Markdown/markdown/issues/1175 as we've already got a good conversation going, but here is fine as well.

facelessuser:
Chunks are handed to `parseChunk`, but a chunk could have multiple blocks separated by `\n\n`. Maybe we should split these before calling `on_add`? This still won't catch cases where a block could be split automatically later:

```
Text
!!! note "Legacy admonition"
    content
```

We allow the same as above for generic blocks, but this would be at least two separate blocks.

```
Text
/// note | Legacy admonition

content
///
```

Andre601:
Regarding the blocks in blocks... I feel like it could become a bit tedious if you for some reason have a lot of nested blocks, to constantly expand the parent by one more `:`

Wouldn't something similar to the tabbed extension be a solution? By that I mean have a specific character to indicate a new box start.

For the sake of it do I use + here as the indicator.
```
::::{info} Block

:::{info} Nested block
Hello!
:::

::::
```
would become
```
:::{info} Block

:::+{info} Nested block
Hello!
:::

:::
```

It's a small change, but I feel like it would be a bit better to have this:
```
:::{info} Extreme nest example

:::+{info} Nested block 1

:::+{info} Nested block 2

:::+{info} Nested block 3

:::+{info} Nested block 4

Hi!

:::

:::

:::

:::

:::
```
...than this
```
:::::::{info} Extreme nest example

::::::{info} Nested block 1

:::::{info} Nested block 2

::::{info} Nested block 3

:::{info} Nested block 4

Hi!

:::

::::

:::::

::::::

:::::::
```

pawamoy:
I'd say it's easier to find the corresponding start/end of a block when they increase/decrease the number of `:`.

2bndy5:
Just looking at your example, I think it is a bit more human readable if the nested blocks have an extra `:` added to their parents. Mainly because I easily get confused which `:::` ends which nested block if it only starts with `:::` or `:::+`. That would be even harder to track when writing lots of content in nested blocks.

facelessuser:
Currently, I'm of the mind of increasing tokens makes things more readable. Nesting will always borderline on becoming less readable, especially when doing deep nesting, but anything to help ease that I'm all for.

I know that `:::` keeps getting referenced, but currently we are using `///` just an FYI. Feel free to voice your opinion about `:::` vs `///`.

The original discussion ([https://github.com/Python-Markdown/markdown/issues/1175](here)) led us to use `///` instead, but other opinions are welcome as well.

2bndy5:
I'm good with `///` because it means less wear-and-tear on my <kbd>shift</kbd> button.

Andre601:
Personally, I'm more in favour of `:::`, mainly because it has established itself as a more or less common syntax for admonitions.
It was even mentioned in a GitHub community discussion multiple times as the "right way to make admonitions"

Regarding the incrementing colons... Would it be not a bit better/easier to look at if it was in reverse order?
Like have childs be increased by 1 compared to parents? Looking at my example, you can see that the most outer start and end are fairly appart and it looks a bit confusing...

> The original discussion ([https://github.com/Python-Markdown/markdown/issues/1175](here)) led us to use `///` instead, but other opinions are welcome as well.

Link is broken. It's /here instead of /1175

2bndy5:
> Would it be not a bit better/easier to look at if it was in reverse order?

I think it would be easier for authors to write this way. And it still maintains readability.

> I'm more in favour of `:::`, mainly because it has established itself as a more or less common syntax for admonitions.

I thought the goal was to break-away from that in favor of a definitive uniformed syntax. If it is modeled after existing syntax, then you risk confusing those who haven't migrated. (see also https://github.com/Python-Markdown/markdown/issues/1175#issuecomment-1222391419)

Andre601:
> I thought the goal was to break-away from that in favor of a definitive uniformed syntax. If it is modeled after existing syntax, then you risk confusing those who haven't migrated. (see also [Python-Markdown/markdown#1175 (comment)](https://github.com/Python-Markdown/markdown/issues/1175#issuecomment-1222391419))

From what I read are most people referencing ::: with admonition, which would be the prime use here...
Like people did mention Pandoc, but also more popular solutions such as Docosaurus use ::: to indicate admonition blocks here, so if anything would it actually help establishing a more common syntax for markdown (if that isn't already the case as some MD renders do use ::: as official admonition syntax from what I know).

2bndy5:
It would be cool if it was configurable, but I imagine the choices for such an option would have to be limited to regex friendly chars.

facelessuser:
I will be releasing a prerelease hopefully sometime soon. The syntax could change for the final release, but I would need a strong majority. I'm personally indifferent to which form gets used `///` vs `:::`. There are some other potentially things that could change as well. I think when I do the prerelease, I will specifically point out some of the things I absolutely want feedback on, but any feedback is of course welcome.

facelessuser:
One thing that will need a decision is whether block specific options should require yaml fencing or have it optional. Optional means we must require a new line between the block header and the content, but mandatory yaml fence means that a new line is not required.

I thought I'd be fine with optional, but I personally disliked it more than I thought I would. I'm leaning towards mandatory. Currently, at least for the prerelease, this will be optional so people can try it and help decide which is better.

2bndy5:
Coming from rST, I'm rather used to putting a blank line between directive options and it's content. On the other hand, I do like having explicit delimiters that tell readers/contributors where the actual content begins and where the options end. If the directive options use YAML syntax, then it makes sense to use fences because YAML is rather easily mistakable for natural text.

facelessuser:
It's not the new line between config and content that's the problem, it's that it isn't easy to tell config from content, so even if you have no config, you have to have a new line.

```
/// note | title

New line required above
///
```

With a distinct fence around the config, you are always sure of when a config exists and can avoid the new line requirement:

```
/// note | title
content
///
```

But maybe this doesn't bother everyone.

2bndy5:
If the title is allowed to span multiple lines in the doc src, then I'm also leaning toward mandatory fencing.

facelessuser:
The "title" does not span multiple lines. That string can be used for anything though, it is dependent on the block:

```
/// tag | div
Div content
///
```

But no, the "title" cannot span multiple lines.

2bndy5:
At the risk of making a pun, I'm on the fence. I'm leaning toward mandatory for readability reasons, but it would probably be best to get my hands dirty first.

I just wrote overrides for Sphinx admonitions in which multi-lined (and optional) titles was a hard feature to support due to docutils' rST parser. We ultimately went with a [new `:title:` option](https://jbms.github.io/sphinx-immaterial/admonitions.html#directive-option-admonition-title) that can concatenate with directive arguments.

facelessuser:
Yeah, that's why I'm making it configurable via the prerelease just to get feedback. I think it may be hard to tell what you'll like or hate until you try it.

2bndy5:
> Would it be not a bit better/easier to look at if it was in reverse order?
> Like have childs be increased by 1 compared to parents

On second thought, I'm not sure I like this idea for parsing reasons. If a block can have children, then the furthest descendant is `///` which I imagine would signify to the parser that the block shall have no children. If it was reversed, then there's nothing in the syntax to prevent the parser from looking for children.

I have an idea I want to try on the eventual RC
```markdown
/// {render-snippet}

/// {note}
Render this code as literal snippet and again as the resulting admonition.
///

///
```
The idea is to bypass having to write demonstrations of MD syntax in duplicates. Currently, the Material theme has to do
~~~markdown
```markdown
::: note
Admonition content
:::
```

<div class="results">

::: note
Admonition content
:::

</div>
~~~
But with a block that accepts no children, the docs' src can be reduced using a custom directive (named `render-snippet` in my example).

facelessuser:
> The idea is to bypass having to write demonstrations of MD syntax in duplicates. Currently, the [Material theme has to do](https://github.com/squidfunk/mkdocs-material/blob/13655d68f5c3514f1a03733ef7b2ecb96bc0d2fe/docs/reference/admonitions.md?plain=1#L116-L137)

It'll probably be possible to some extent:

```
//// render-snippet

/// note
Render this code as literal snippet and again as the resulting admonition.
///

////
```

You could store the content in a raw code block, and then using an `on_end` event extract that text and run it back through the parser for the results. You always run the risk of not catching the content before changes if a preprocessor plugin got to it first as these generic Blocks run as a blockprocessor which runs after preprocessor. We do at lest unwind what SuperFences does if saving content as "raw" text.

I do use a [special custom fence](https://github.com/facelessuser/pymdown-extensions/blob/main/tools/pymdownx_md_render.py) in SuperFences that allows me to execute and render Markdown in an isolated sandbox. It allows me to demonstrate Markdown using different options as the content is isolated:

````
```md-render
---
extensions:
- pymdownx.magiclink
- pymdownx.saneheaders
- markdown.extensions.attr_list

extension_configs:
  pymdownx.magiclink:
    repo_url_shortener: true
    repo_url_shorthand: true
    social_url_shorthand: true
    social_url_shortener: true
    user: facelessuser
    repo: pymdown-extensions
---
- Just paste links directly in the document like this: https://google.com.
- Or even an email address: fake.email@email.com{.magiclink-ignore}.
```
````

There is some limitations of course. As it is isolated, you have to be careful as you could generate conflicting IDs, and some preprocessing could get a hold of the text before you can process it. I don't generate both the source in a code block and then the results, in this case, I just render the results, but what you describe could be done with custom fences.

**Comment by pawamoy** [comment]
(I've implemented a similar fence that allows to render both the source and result, see https://pawamoy.github.io/markdown-exec))

**Comment by facelessuser** [comment]
Yep, I keep forgetting, but https://pawamoy.github.io/markdown-exec/ does probably already offers what you need in an already-packaged way. There is quite a lot you can do with custom fences. For example, I also created interactive playgrounds that translate code into REPL outputs and preview colors. Then we load Python in the browser (using Pyodide) and allow the viewer to interact and execute code on the fly.

Basically, there is a lot of potential already available via custom fences and SuperFences, and `markdown-exec` seems to leverage some of that already. Custom fences also capture the content pretty early in a pre-processor.  Generic blocks will capture a bit later in the process but allow for translating and wrapping Markdown into more HTML constructs.

**Comment by 2bndy5** [comment]
Good to know. Thanks. `markdown-exec` does seem to offer [exactly what I wanted](https://pawamoy.github.io/markdown-exec/usage/#literate-markdown) to do.

**Comment by facelessuser** [comment]
Both `:::` vs `///` syntax and mandatory vs optional YAML config fences will be configurable during the alpha/beta stage so that people can try both and help give feedback as to the direction we should take.

File: docs/src/markdown/extensions/blocks/index.md [hunk 1]
```diff
@@ -0,0 +1,199 @@
...
+   authors who'd like to not have so many nested indentation levels in the documentation. Additionally, most code
+   editors will syntax highlight such nested constructs as indented code blocks which can make reading the source
+   difficult.
+
+Blocks is a plugin that essentially allows for the creation of its own meta-plugins that function as fenced containers.
+Each meta-plugin can do whatever it wants with the content inside the container and allow for a single generic format
+to create as many different block style plugins as desired. As the blocks are fenced, indentation is not required to
+determine the start and end of the blocks.
+
+Blocks also allows for per block options given the extension authors an easy way to extend functionality and users an
```
pawamoy, line 31:
```suggestion
Blocks also allows for per block options giving the extension authors an easy way to extend functionality and users an
```

docs/src/markdown/extensions/blocks/index.md
```diff
@@ -0,0 +1,199 @@
...
+import markdown
+md = markdown.Markdown(extensions=['pymdownx.blocks'])
+```
+
+## Feedback Requested
+
+While any and all feedback is welcome, there are two undecided options that require feedback to help us finalize the
+syntax for generic blocks. During the alpha/beta stage, we provide global options so users can try out the different
+options and provide their own feedback. Feedback will be taken into consideration as we move towards an official release
+candidate. It will be impossible to make all users happy, but we we'd like to give users an opportunity to help shape
```
pawamoy, line 46:
```suggestion
candidate. It will be impossible to make all users happy, but we'd like to give users an opportunity to help shape
```

docs/src/markdown/extensions/blocks/index.md
```diff
@@ -0,0 +1,199 @@
...
+                }
+            }
+        }
+    }
+)
+```
+
+## Syntax
+
+Syntax for Blocks requires the desired content to be included between fences. Fences are denoted by by using three or
```
pawamoy, line 131:
```suggestion
Syntax for Blocks requires the desired content to be included between fences. Fences are denoted by using three or
```

facelessuser:
The first alpha will be focused on syntax feedback. I'm not going to bother documenting the API or anything initially. It will be more to give people a chance to get a feel for the generic blocks. It will implement admonitions, details, tabs, and arbitrary HTML.

pawamoy:
Thinking about graceful degradation (rendering blocks where they are not supported, for example on GitHub), I think enforcing a new line can be useful. Consider these two examples and how they would render without Blocks support:

### 1.

```
/// note
A note about something.
///
```

Rendered as:

/// note A note about something. ///

### 2.

```
/// note

A note about something.
///
```

Rendered as:

/// note

A note about something. ///

---

I believe the second example is easier to read, i.e. the degradation is more graceful.

Just a thought :smile:

facelessuser:
> I believe the second example is easier to read, i.e. the degradation is more graceful.

Certainly, something to think about. I'll be honest, GitHub degradation is not the highest bullet point on my list, what will be most interesting is how it feels in your documentation. I think once people start playing with it, such things may become more clear.

Andre601:
I'm curious about something...
With this change, would this also have an effect on other extensions like the details one?
I can see from the more recent changes that tabs get a special type added here to use? Will that be the same for details later on (I assume so)?

facelessuser:
@Andre601 I'm not sure I quite understand the question, can you elaborate?

facelessuser:
Maybe this is what you mean. Admonitions have a `type` option. This is just a quick, easy way to set the class, just like the original admonition extensions. Details allows something similar, and when you do add a class, it can operate as a collapsible admonition (assuming it is styled as such). MkDocs Material utilizes this, so I decided to treat details in a similar way moving forward so various types could be registered as a shortcut.

With that said, I have no intention on adding a `type` option to blocks where it doesn't make sense. This doesn't make sense for things like Tabs, so I have no intention to add a type for tabs or for HTML blocks. If a class or classes are desired, they can be added with the `attributes` option.

Andre601:
> @Andre601 I'm not sure I quite understand the question, can you elaborate?

Right now you have `??? <type> "<title>"` for inserting details using the details extension. Will this one be updated too to fit the changes made here?

> Maybe this is what you mean. Admonitions have a `type` option. This is just a quick, easy way to set the class, just like the original admonition extensions. Details allows something similar, and when you do add a class, it can operate as a collapsible admonition (assuming it is styled as such). MkDocs Material utilizes this, so I decided to treat details in a similar way moving forward so various types could be registered as a shortcut.
> 
> With that said, I have no intention on adding a `type` option to blocks where it doesn't make sense. This doesn't make sense for things like Tabs, so I have no intention to add a type for tabs or for HTML blocks. If a class or classes are desired, they can be added with the `attributes` option.

That answers it I guess.
I was asking because I saw you added examples for tabs (replacement for tabbed extension) in one of the markdown files (https://github.com/facelessuser/pymdown-extensions/pull/1777/files#diff-a094eddcb595c258969637e9855a7cc50d64abb7c86b65436d57e574a8948a58) so I wanted to make sure that I don't missunderstand things.

facelessuser:
> Right now you have `??? <type> "<title>"` for inserting details using the details extension. Will this one be updated too to fit the changes made here?

What's to update? They are compatible but different extensions. The old details already supports classes/type:

```
??? note "title"
     content
```

The new general block extension now has a details plugin that can do the same:

```
/// details | title
---
type: note
---

content
///
```

If you register a  special type for details, then you can do something like shown below. Let's say we registered a type `details-type`, now we can do this:

```
/// details-type | title

content
///
```

sisp:
:+1: on this feature!

Another advantage of this syntax is compatibility with, e.g., [`remark-directive`](https://github.com/remarkjs/remark-directive) which enables using [`remark-lint`](https://github.com/remarkjs/remark-lint) (perhaps with additional linter plugins if necessary, not sure yet) for linting Markdown. `!!! class`-style admonitions are incompatible with `markdownlint` and `remark-lint` at the moment and cause confusion with indented code blocks syntax.

facelessuser:
> Another advantage of this syntax is compatibility with, e.g., [`remark-directive`](https://github.com/remarkjs/remark-directive) which enables using [`remark-lint`](https://github.com/remarkjs/remark-lint) (perhaps with additional linter plugins if necessary, not sure yet) for linting Markdown. `!!! class`-style admonitions are incompatible with `markdownlint` and `remark-lint` at the moment and cause confusion with indented code block syntax.

I assume you are commenting on the original syntax as mentioned near the start of the thread. I think we've moved a bit more away from replicating the syntax of things like `remark-directive` as we do not plan to have 100% compatibility with directives per se, but are more inspired by directives.  With that said, `!!! class` is definitely not very compatible with `markdownlint`, and this fact is definitely a driving factor in this new generic block syntax.

I am hoping to get an alpha out today allowing people to test drive and give feedback relating to the syntax and possible syntax directions we can take. We do have a couple of undecided directions that will have feature switches during the alpha period to let people try them and then give feedback. I hoped to get it out yesterday, but stuff got in the way, but I think we can get an alpha out today.

After getting more feedback on syntax, I'll probably document the API and make any last changes (if any) and then release another alpha to get feedback on the API when creating custom generic blocks.

Once everything is settled, we'll probably go into beta to help shake out any unexpected bugs.

Keep in mind that the implementation is subject to change.

oprypin:
For admonitions compatible with `markdownlint` there is already https://github.com/oprypin/markdown-callouts/ - and these also still look natural in fallback vanilla Markdown rendering.

facelessuser:
Yes https://github.com/oprypin/markdown-callouts is available, and if people are looking for that as a specific solution, it is definitely a good alternative. With that said, I'd like to not get off on a tangent topic, especially in a PR for a specific feature.

sisp:
> I assume you are commenting on the original syntax as mentioned near the start of the thread.

Ah, that's right. It seems I hadn't read the discussion carefully that the proposed syntax has changed.

> I think we've moved a bit more away from replicating the syntax of things like remark-directive as we do not plan to have 100% compatibility with directives per se, but are more inspired by directives.

To be honest, I'm a bit sad to read that because a non-standard (or even non-"official"-proposal-based) syntax makes it very cumbersome to use standard tools like `remark-lint` for linting Markdown. The `!!! class`-based admonitions syntax (and other non-standard extensions) are already blockers in that regard and creating `remark` extensions for especially this indentation-based content block is not entirely trivial because AFAIK it needs a `micromark` extension and a `mdast` extension before a proper linter rule can be created. In my opinion, it would be much better if Markdown extensions like admonitions used at least the the [CommonMark generic directives proposal](https://talk.commonmark.org/t/generic-directives-plugins-syntax/444) for which tool support exists.

Andre601:
IF the syntax stays as `///` (given that people prefer it over `:::`) do I really hope that there will be an option to change it to other formats... But that's what the alpha will be about anyways, so let's see how it goes...

facelessuser:
> Ah, that's right. It seems I hadn't read the discussion carefully that the proposed syntax has changed.

There are still things that are up in the air, whether we use `:::` or `///` and things like that. Once we get it into people's hands, we'd definitely like to get feedback on whatever.

> To be honest, I'm a bit sad to read that because a non-standard (or even non-"official"-proposal-based) syntax makes it very cumbersome to use standard tools like remark-lint for linting Markdown.

Right, I should state, that our main objective is to fix the problematic case where indented blocks are often treated wrongly by editors and other things, not necessarily to make it work in things like `remark-lint`.

Keep in mind that Python Markdown has been around long before CommonMark existed. It is not a CommonMark parser and moving it to be a CommonMark parser would require a full rewrite. These are plugins for Python Markdown, not a CommonMark solution.

We wanted a generic block case, and sure we could make it more like some other solutions out there, but we are not aiming for 1:1 feature parity. Some people may prefer `:::` or find `///` easier to look at, that is why I'm trying to get feedback.

There currently is no definitive, official solution that I'm aware of right now for Markdown.

facelessuser:
Python Markdown will likely never be rewritten. This is just a statement from my experience. I understand the desire people have for all Markdown everywhere to be CommonMark. I do not control Python Markdown, but it isn't CommonMark and doubt it will ever be so.

There are other CommonMark compliant Python libraries out there, so having yet another may not make sense. Regardless, Python Markdown is a very flexible Markdown implementation which is why we have these plugins. It is "old school" in its Markdown approach, and it certainly isn't perfect, but it is what is.

That's probably all I really want to say about that as I don't want to get off on more tangents.

facelessuser:
I have released an alpha containing the current changes: https://pypi.org/project/pymdown-extensions/#history.

Please provide feedback in this discussion: https://github.com/facelessuser/pymdown-extensions/discussions/1868

facelessuser:
@squidfunk if you are interested, the first alpha is released. Feel free to make a note in Material for any and all interested in giving feedback to help shape the feature. Feel free also to try it out yourself and provide feedback. See the post above for the discussion for feedback.

facelessuser:
FYI, alpha 2 discussion here: https://github.com/facelessuser/pymdown-extensions/discussions/1883

facelessuser:
Need to revise documentation. We are going to make the config use indentation by default for now. I think we may remove the leading character (blockquote style) config.

facelessuser:
@gir-bot lgtm

PR #1811, opened 2022-09-16
Tab IDs are slugified from tab title

### LINKED ISSUE #1807: Allow slugify function for Tabbed to generate readable anchors
### Description

Currently, `ids` generated by Tabbed are of the form `__tabbed_{tabset}_{tab}`, e.g. `__tabbed_1_2`. Material for MkDocs allows to [add anchor links to tabs](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#anchor-links), which means you can now explicitly link to tabs. However, this generates rather ugly and non-permalink URLs: https://squidfunk.github.io/mkdocs-material/getting-started/#__tabbed_1_2

### Benefits

[Users asked](https://github.com/squidfunk/mkdocs-material/discussions/4360) whether the ids can be customized, and immediately the Attribute Lists extension popped up in my head. However, we probably don't need support for Attribute Lists, as the usage of a user-definable slug function would probably be enough. The slug function could be passed the contents of the tab label, and generate a readable anchor.

### Solution Idea

For example, with this Markdown:

```
=== "This is a label"

    Content
```

The following id could be generated, which can be used as an anchor:

``` html
...
<input id="this-is-a-label" name="__tabbed_1" type="radio">
...
```

The [Table of Contents](https://python-markdown.github.io/extensions/toc/#usage) extension also implements this, which is related, as it also generates anchor links. Configuration could possibly look like this:

``` yaml
markdown_extensions:
  pymdownx.tabbed:
    slugify: !!python/name:pymdownx.slugs.uslugify
```

And as a shortcut defaulting to the standard Markdown slugify function:

``` yaml
markdown_extensions:
  pymdownx.tabbed:
    slugify: true
```

facelessuser:
Resolves #1807

### DIFF
--- a/pymdownx/tabbed.py
+++ b/pymdownx/tabbed.py
@@ -23,6 +23,8 @@ DEALINGS IN THE SOFTWARE.
 """
 from markdown import Extension
 from markdown.blockprocessors import BlockProcessor
+from markdown.treeprocessors import Treeprocessor
+from markdown.extensions import toc
 import xml.etree.ElementTree as etree
 import re
 
@@ -35,14 +37,15 @@ class TabbedProcessor(BlockProcessor):
     )
     COMPRESS_SPACES = re.compile(r' {2,}')
 
-    def __init__(self, *args, alternate_style=False):
+    def __init__(self, parser, config):
         """Initialize."""
 
-        super().__init__(*args)
+        super().__init__(parser)
         self.tab_group_count = 0
         self.current_sibling = None
         self.content_indention = 0
-        self.alternate_style = alternate_style
+        self.alternate_style = config['alternate_style']
+        self.slugify = callable(config['slugify'])
 
     def detab_by_length(self, text, length):
         """Remove a tab from the front of each line of the given text."""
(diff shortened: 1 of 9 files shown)


facelessuser:
Tests still need to be written.

pymdownx/tabbed.py
```diff
@@ -288,14 +291,66 @@ def run(self, parent, blocks):
...
+                    else:
+                        for i in list(el):
+                            if i.tag == 'input':
+                                inputs.append(i)
+                            if i.tag == 'label':
+                                labels.append(i)
+
+                    # Generate slugged IDs
+                    for inpt, label in zip(inputs, labels):
+                        # Do not override pre-existing ids
```
facelessuser, line 336: This comment is unnecessary as we are not actually doing anything to avoid pre-existing ids. There currently isn't even an easy way to specify manual IDs so there is no reason to expect them.

facelessuser:
Documentation is also required

facelessuser:
@gir-bot lgtm

PR #1822, opened 2022-09-29
Add support for specifying specific line numbers in Snippets

### LINKED ISSUE #1462: Allow snippets to only show certain user defined lines from the included file
### Description

When using `literalinclude` in a rst file you can specify the lines from the included file that are shown, e.g.:

```rst
.. literalinclude:: example.py
   :lines: 1,3,5-10,20-
```

Would it be possible to include something similar for snippets, using, e.g.:

```md
-->8-- "example.py" lines: 1,3,5-10,20-
```

### Benefits

This would allow users to only show required parts of a file, e.g., just show one function from a larger Python file.

### Solution Idea

My regular expression knowlegde isn't good enough to know how to add this into `RE_ALL_SNIPPETS`, but if the line numbers can be read and converted into a list of lines to include (in, say, a variable called `includelines`), then I assume:

```python
with codecs.open(snippet, 'r', encoding=self.encoding) as f:
    new_lines.extend(
        [space + l2 for l2 in self.parse_snippets([l.rstrip('\r\n') for l in f], snippet)]
    )
```

could be changed to:

```python
with codecs.open(snippet, 'r', encoding=self.encoding) as f:
    snippetlines = f.readlines()
    new_lines.extend(
        [space + l2 for l2 in self.parse_snippets([snippetlines[i].rstrip('\r\n') for i in includelines], snippet)]
    )
```

facelessuser:
Resolves #1462

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -3,6 +3,7 @@
 ## 9.6
 
 - **NEW**: Add options for generating tab IDs from tab titles.
+- **NEW**: Add support for specifying specific lines for Snippets.
 
 ## 9.5
 
--- a/docs/src/markdown/extensions/snippets.md
+++ b/docs/src/markdown/extensions/snippets.md
@@ -100,6 +100,16 @@ include.md
 ; skip.md
 --8&lt;--</code></pre>
 
+## Snippet Lines
+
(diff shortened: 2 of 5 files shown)


facelessuser:
@gir-bot lgtm

PR #1825, opened 2022-10-01
Allow specifying a section in a snippet

### LINKED ISSUE #1824: Snippets : warn against file or part changes through checksum
### Description

Issue #1462, closed not long ago, added a feature to allow including only specified line ranges from external file with snippets. 
One concern regarding this feature was inconsistency it could create if file changed (e.g if lines were added/removed, creating an offset).  
I think a checksum mechanism could be used to warn against changes, and it isn't limited to range inclusion, but could also benefit general snippets.

## Context

I use pymdownx through Material for Mkdocs to build a computer science course. The course is for absolute beginners, and quite young. For this reason, there are heavy constraints :
 
1. I can't give them full code if the file is large, otherwise they get confused, as for example if tests are included.
2. I often need to explain the code in text, and thus make references to lines.
3. I refer to some snippets at different places, for example exercises are often referred to in the course and in an "exercises" part of the doc.

Range inclusion solves problem 1, but inconsistency is still a danger here, as even in file fully included, an inconsistency in lines numbers can confuse the students. 

I think a lot of tutorials and docs have this problems to a certain extent, and it would be beneficial to have a change warning mechanism.

### Benefits

Protects against possible inconsistency when a file included through snippets in the docs changes, makes easier to structure large docs with multiple references to the same file, or with references to out-of-docs files, as test cases.

### Solution Idea

Here is a draw at a possible solution : 

Through the inclusion syntax, snippets could allow user to provide, at the end of file name chain, a checksum between bracket. So for example : 

``somefile::4[EA8A25E440]``

Then snippet extension would check this checksum against either : 
- The entire file (likely easier to implement, but a slight bit less practical, yet sufficient i think)
- The snippet included (more practical, but maybe more difficult to implement) 

In case a checksum isn't matching, a warning message would be issued, mentioning both checksum : 

```
Inclusion somefile::4[EA8A25E440] has mismatching checksum for somefile::4 : expected EA8A25E440, yet found EA5E4408A2.
```

This way, it is simple for user to update checksum after checking file content. Warning would likely be failing the build, or maybe an option in config could allow to choose whether warn or fail.

An empty checksum `[]` would be considered as a checksum as well, always failing, simply to allow user to use snippets to compute checksum.

facelessuser:
Related #1824

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -4,6 +4,7 @@
 
 - **NEW**: Tabbed: Add new syntax to allow forcing a specific tab to be select by default.
 - **NEW**: Snippets: Add new option to pass arbitrary HTTP headers.
+- **NEW**: Snippets: Allow specifying sections in a snippet and including just the specified section.
 
 ## 9.6
 
--- a/docs/src/markdown/extensions/snippets.md
+++ b/docs/src/markdown/extensions/snippets.md
@@ -120,6 +120,44 @@ include.md::3
 ;--8<--
 ```
 
+### Snippet Sections
+
+!!! new "New 9.7"
+
+Specifying snippet lines may not always be as the source could change moving, adding, and/or removing lines. A way
+around this is to partition a snippet into named sections and then targeting a specific section to be included instead
+of specific line numbers.
+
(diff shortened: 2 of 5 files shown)


facelessuser:
@gir-bot lgtm

PR #1834, opened 2022-10-19
Add support for url headers

michalkowol:

PR #1836, opened 2022-10-20
Add support for url headers (tests)

michalkowol:


michalkowol:
https://github.com/facelessuser/pymdown-extensions/pull/1834#issuecomment-1285709265
I think it would be nice to have additional test to check url and headers
```python
assert mock_urlopen.call_args.args[0]._full_url == expected_url
assert mock_urlopen.call_args.args[0].headers == expected_headers_dict
```

tests/test_extensions/test_snippets.py
```diff
@@ -690,3 +691,42 @@ def test_missing(self, mock_urlopen):
...
+            R'''
+            --8<-- "https://test.com/myfile.md"
+            ''',
+            '''
+            <p>contents</p>
+            ''',
+            True
+        )
+        assert mock_urlopen.call_args.args[0]._full_url == 'https://test.com/myfile.md'
+        assert mock_urlopen.call_args.args[0].headers == {'User-agent': 'Mozilla/5.0', 'X-foo': 'Foo'}
```
michalkowol, line 732: https://github.com/facelessuser/pymdown-extensions/pull/1834#issuecomment-1285709265
I think it would be nice to have additional test to check url and headers
```python
assert mock_urlopen.call_args.args[0]._full_url == expected_url
assert mock_urlopen.call_args.args[0].headers == expected_headers_dict
```

facelessuser:
I'm kind of indifferent about the test only because we aren't testing our logic, but testing that urllib processes the URL stuff properly. That's why we mock it, because we aren't interested in ensuring their library works (that's their job) as much as we are testing that we handle what it returns properly.

tests/test_extensions/test_snippets.py
michalkowol: There is a small problem with Python 3.7 - checking...

michalkowol:
> I'm kind of indifferent about the test only because we aren't testing our logic, but testing that urllib processes the URL stuff properly. That's why we mock it, because we aren't interested in ensuring their library works (that's their job) as much as we are testing that we handle what it returns properly.

I think you are right! I am closing this PR! Thanks!

PR #1866, opened 2022-12-01
Fix snippet sections introduction

StevenMaude:


facelessuser:
Good catch!

@gir-bot lgtm

PR #1875, opened 2022-12-09
Add tip to avoid trailing whitespace

scottamain:
Because a whitespace is required for fenced code blocks on the first line of a list item, it's a better practice to include an explicit whitespace so it does not get accidentally removed. (Using `&nbsp;` does not have the same effect because that character gets parsed into the HTML as a character string so you end up with a blank line before the code block, whereas the unicode space gets collapsed and ignored by the Markdown parser.)

### DIFF
--- a/docs/src/markdown/extensions/superfences.md
+++ b/docs/src/markdown/extensions/superfences.md
@@ -58,17 +58,17 @@ md = markdown.Markdown(extensions=['pymdownx.superfences'])
     ````
 
 7. If using a fenced block as the first line of a list, you will have to leave the first line blank, but remember that
-  the list marker must be followed by a space. To avoid the trailing whitespace that might be removed by your IDE,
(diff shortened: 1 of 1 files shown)


facelessuser:
Forgive me, but I'm confused about what this is trying to convey:

```
-<space>&#32;
```

Are you suggesting you could use `&#32;` in place of a space?

`<space>` is simply meant to convey a space, it could be `&#32;` if desired, but I find `-<space>&#32;` confusing. If you'd like to add a little admonition tip like below, I could add that:

```
!!! tip "Spaces"
    `<space>` is simply meant to represent a white space character, it can be a literal space or an HTML entity such as `&#32;` for better visibility.
```

facelessuser:
Okay, yeah, I think I like that better. It's less busy and I like the updated explanation better as well.

facelessuser:
@gir-bot lgtm

PR #1882, opened 2022-12-15
FIX: smartstong -> smartstrong

perpil:
Fixes minor typo.


facelessuser:
Good catch, thanks!

@gir-bot lgtm

PR #1904, opened 2023-01-10
Replace MaxCDN, which has shut down

samdobson:
Twemoji is currently broken, due to MaxCDN shutting down. This may only be a short-term fix.

See https://github.com/twitter/twemoji/issues/580


facelessuser:
I do imagine this will break some tests.

samdobson:
My bad. I think this should do it.

facelessuser:
I imagine from the root directory you will have to run `python3 run_tests.py -u`.

samdobson:
Thanks for the pointer. Tests are now passing locally. I force-pushed to clean up the mess.

facelessuser:
Can you install Pygments locally and re-run `run_tests.py -u`? We run with Pygments installed.

samdobson:
No worries, that's done. Not too sure what has made codecov unhappy.

facelessuser:
Codecov isn't unhappy. Recently, there is a lag for things to get updated their, or sometimes certain tests don't get updated at all. It's kind of annoying.

PR #1917, opened 2023-01-20
Relax Snippets syntax to allow only one dash `-8<-`

### LINKED ISSUE #1916: Import Snippet Section from XML file
### Discussed in https://github.com/facelessuser/pymdown-extensions/discussions/1914

<div type='discussions-op-text'>

<sup>Originally posted by **maciejwalkowiak** January 19, 2023</sup>
I want to use Snippet Sections in an XML file:

```xml
<dependencies>
    <!-- --8<-- [start:func] -->
    <dependency>
        <groupId>software.amazon.awssdk</groupId>
        <artifactId>dynamodb-enhanced</artifactId>
    </dependency>
    <!-- --8<-- [end:func] -->
</dependencies>
```

`--8<--` within an XML comment makes the XML invalid which as a result breaks the XML tooling. The error is following:

```
in comment after two dashes (--) next character must be > not 8 (position: START_TAG seen ...<dependencies>\n        <!-- --8... @21:17) 
```

Is there a way to use different "tag" than `--8<--` or any other way to make snippet sections work with XML files?</div>

facelessuser:
Resolves #1916

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 9.9.2
+
+- **FIX**: Snippets syntax can break in XML comments as XML comments do not allow `--`. Relax Snippets syntax such that
+  `-8<-` (single `-`) are allowed.
+
 ## 9.9.1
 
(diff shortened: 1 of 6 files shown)


facelessuser:
@gir-bot lgtm

PR #1926, opened 2023-01-28
Update year to 2023

reregaga:


facelessuser:
Thanks for this!

I would update the mkdocs.yml in docs/src/mkdocs.yml, the one you updated will get overwritten when we update assets. When we package our custom assets, we use the one in docs/src/mkdocs.yml as the template and replace the settings pointing to the old assets with the new assets.

facelessuser:
@gir-bot lgtm

PR #1963, opened 2023-02-22
Snippet enhancements

### LINKED ISSUE #1945: Unindent for snippets
### Description

It would be nice if snippets would be unindented, so that the leading whitespace that is common to all lines in the snippet is removed, so it starts left-most with the first character.
This should imho even be the default behavior if it will be configurable.

### Benefits

No need to cram the snippet code to beginning of line in the source file to have it nice in the docs.

### Solution Idea

Just get the leading whitespace of all lines, take the shortest one and remove that from all lines.
If someone has mixed space and tab indents, it is his own fault. :-D

### LINKED ISSUE #1946: Fail on missing snippet sections
### Description

To prevent typos or accidental section deletions, missing snippet sections should be complained about, either always, or configurable, or together with complaining about missing files option.

### Benefits

Prevent typos, accidential deletions, ...

### Solution Idea

_No response_

### LINKED ISSUE #1947: Snippet IDs with dashes and underscores
### Description

I tried to use snippet IDs with dashes and / or underscores.
Unfortunately it lead to the whole snippet file not being found (build failing with `SnippetMissingError`).
So it seems currently only camelCase IDs are supported, but neither kebab-case nor snake_case.
Would be nice to add some more flexibility here.

### Benefits

For such IDs I usually prefer kebab-cased IDs for readability.

### Solution Idea

_No response_

facelessuser:
- Allow `-` and `_` in section names
- If `check_paths` is enabled and a section cannot be found, raise an error.
- Add `dedent_subsections` to remove indentation of sections or lines that are inserted.

Resolves #1947
Resolves #1946
Resolves #1945

### DIFF
--- a/docs/src/dictionary/en-custom.txt
+++ b/docs/src/dictionary/en-custom.txt
@@ -20,6 +20,8 @@ CommonMark
 CriticMarkup
 Ctrl
 DOM
+De
+Dedent
 Dicts
 Donath
 ElementTree
@@ -138,6 +140,8 @@ config
 configs
 coveragepy
 customizable
+de
+dedenting
 deprecations
 dev
 deviantart
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -2,11 +2,16 @@
 
 ## 9.10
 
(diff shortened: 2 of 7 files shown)


facelessuser:
@gir-bot lgtm

PR #1964, opened 2023-02-23
Relax hash requirement in link shortening and update GitLab links

### LINKED ISSUE #1895: Add MagicLink support for abbreviated commit IDs
### Description

Thanks for maintaining these extensions! I have a question about autolinking commit IDs with the MagicLink extension.

I'm aware that [MagicLink already supports autolinking the full 40 character commit IDs](https://facelessuser.github.io/pymdown-extensions/extensions/magiclink/#commits) (SHAs). It recognizes these commit IDs with regexes:

https://github.com/facelessuser/pymdown-extensions/blob/25508d4ea52b46c028b37a3cef09555d4f61a876/pymdownx/magiclink.py#L118-L122

MagicLink converts these full commit IDs into an abbreviated format. The length of these abbreviated commit IDs is the first 7 characters on GitHub and BitBucket, and the first 8 characters on GitLab, as shown in each `hash_size` item in the `PROVIDER_INFO` dictionary constant.

https://github.com/facelessuser/pymdown-extensions/blob/25508d4ea52b46c028b37a3cef09555d4f61a876/pymdownx/magiclink.py#L215

https://github.com/facelessuser/pymdown-extensions/blob/25508d4ea52b46c028b37a3cef09555d4f61a876/pymdownx/magiclink.py#L436-L439

So, MagicLink creates links from 40 character commit IDs. **Could MagicLink also create links from abbreviated commit IDs?**

### Benefits

Users would be able to link to individual commits by specifying abbreviated commit IDs, which would be helpful when creating changelogs.

The autolinking behavior of MagicLink would more closely match the behavior of platforms like GitHub. For example, [GitHub will autolink](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls) the commit ID below, even though it's only the first 7 characters `25508d4`:

25508d4

And note that the full commit ID is not needed to construct the link. Navigating to https://github.com/facelessuser/pymdown-extensions/commit/25508d4 points to the same commit.

### Solution Idea

Autolink abbreviated commit IDs, based on the `provider:`, `user:`, and `repo:` settings passed in.

Using the example above, and the following options:

- `provider`: `'github'` (the default)
- `user`: `facelessuser`
- `repo`: `pymdown-extensions`
- `repo_url_shorthand`: `True`

MagicLink would transform `25508d4` into `https://github.com/facelessuser/pymdown-extensions/commit/25508d4`.

It seems to me that the main challenge would be parsing the abbreviated commit ID.

It's more challenging than the full 40 character ID, because there are few words that are 40 characters, but many more words that are 7 characters, so there's a risk of false positives. Because of the risk of false positives, I would recommend disabling autolinking of abbreviated commit IDs by default, and offering a Boolean [configuration option](https://facelessuser.github.io/pymdown-extensions/extensions/magiclink/#options) to enable it. Maybe something like `link_abbreviated_commit_ids`.

facelessuser:
Resolves #1895


facelessuser:
@gir-bot lgtm

PR #1966, opened 2023-02-25
Fix block modes

facelessuser:
- inline should account for children in the block
- raw should unescape all content in stash.
- HTML should use the same block mode names as Blocks.


facelessuser:
@gir-bot lgtm

PR #1991, opened 2023-03-17
Corrected a typo: "be" missing

arterm-sedov:
Original text: There _may extensions_ that use names after hashes
Corrected text: There may **be** extensions that use names after hashes


facelessuser:
@gir-bot lgtm

Issue #1980, opened 2023-03-10
[Feature request] Support nested snippets by excluding all scissor lines from rendering


ari-noumena:
### Description

When writing tutorials, a common use-case is to first show small parts of a file, before showing more extensive parts (or the entire file). For example:

```
// --8<-- [start:hello]
fun hello() {
	blah blah;

	// --8<-- [start:interesting]
	// Do some interesting stuff here!
	var x = foo();
	bar(x);
	// --8<-- [end:interesting]
}
// --8<-- [end:hello]
```

When `hello` is then included in some markdown, it will render a block like so:

```
fun hello() {
	blah blah;

	// --8<-- [start:interesting]
	// Do some interesting stuff here!
	var x = foo();
	bar(x);
	// --8<-- [end:interesting]
}
```

But the scissors aren't at all relevant to the example, and it's not desirable for the user to see them. It would be much nicer if they would see the following:

```
fun hello() {
	blah blah;

	// Do some interesting stuff here!
	var x = foo();
	bar(x);
}
```

One workaround is of course to *duplicate* the code (either in the same or different files, depending on how it works) such that there are no nested snippets, but that's a bit messy and makes it harder to maintain.

### Benefits

It would benefit anyone wanting to render snippets containing nested snippets.

### Solution Idea

If there are reasons for not wanting this behavior by default, perhaps it could be enabled as an option.

facelessuser:
This seems like a reasonable request. I'll have to look and see if there are any complications, but I can't seem to think of any off the top of my head. I imagine we can do it as a post-processing step once the entire nested snippet is acquired. Just remove lines with the snippet syntax.

PR #2002, opened 2023-04-04
Add option to omit scissor lines from snippet extractions

ari-noumena:
This is my attempt to implement #1980. I thought it would be best to add as a disabled-by-default feature.


facelessuser:
@ari-noumena Thank you. I actually already have this working locally, I just got side tracked on some of the other projects I support. I'll take a look though and see if there is any thing extra here that makes sense to pull in.

PR #2003, opened 2023-04-04
Ignore nested snippet sections

### LINKED ISSUE #1980: [Feature request] Support nested snippets by excluding all scissor lines from rendering

facelessuser:
Resolves #2002 
Resolves #1980

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 9.11
+
+- **NEW**: Snippets: Ignore nested snippet section syntax when including a section.
+
 ## 9.10
 
 - **NEW**: Blocks: Add new experimental general purpose blocks that provide a framework for creating fenced block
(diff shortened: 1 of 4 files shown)


facelessuser:
@gir-bot lgtm

PR #2023, opened 2023-04-24
Use correct name for PyYAML dependency

eliroca:


facelessuser:
Installation via pop is case insensitive. Additionally, this is the casing used in the install example in the official documentation of PyYAML.

eliroca:
Thanks, I misinterpreted a warning from RPMLINT.. 
```
[   39s] RPMLINT report:
[   39s] ===============
...
python39-pymdown-extensions.noarch: W: python-missing-require pyyaml
[   40s] The python package declare some requirement that's not detected in the rpm
[   40s] package. Please, verify that all dependencies are added as Requires.
```

facelessuser:
Yeah, this seems to be an issue with their packaging of our lib. We officially support installation via pip. If there are issues with other package managers, the issue should be brought to their attention.

PR #2039, opened 2023-05-16
Fix regression of snippets nested deeply under specified base path

### LINKED ISSUE #2038: pymdown-extensions 10.0 seems to break subdirectory imports
### Description

I've got the following file structure (for clarity, really - to keep imported files clearly separated)

```
docs/                                                                                                                                       
├── parent-file.md                                                                                                                 
├── includes
│   ├── includeA.md
│   ├── includeB.md
└── index.md

```

with the following configuration:

```
- pymdownx.snippets:
      base_path: docs
      check_paths: true
```

This config used to work fine until the update of 10.0 - which i understand will restrict imports to relative paths.

```
...
    base_path: docs
```

in the file `parent.md` - 
```
--8<-- "includes/includeA.md"
...
--8<-- "includes/includeB.md"

```

My path is relative to the parent directory, and contained in the parent directory (which shouldn't be a vulnerability).(everything is within the `./docs/` directory) - yet imports fail.

I've tried several possibilities for base_path (`./docs`, `./docs/`,  `docs/`) - but the only possibility that seems to work is to drop the sub-directory component and change the imports 

```
...
    base_path: docs/includes
```

```
--8<-- "includeA.md"
...
--8<-- "includeB.md"

```

Now i would prefer to keep the original structure for clarity, and to avoid problems when other files are imported (imports are with `imports/includeA.md`) but i'd also not want to enable legacy mode, as in my understanding, that will reenable the vulnerability.

### Version(s) & System Info

- Operating System: linux
- Python Version: 3.10
- Package Version: 10.0

facelessuser:
Fixes #2038

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 10.0.1
+
+- **FIX**: Regression related to snippets nested deeply under specified base path.
+
 ## 10.0
 
 - **Break**: Snippets: snippets will restrict snippets to ensure they are under the `base_path` preventing snippets
--- a/pymdownx/__meta__.py
+++ b/pymdownx/__meta__.py
@@ -185,5 +185,5 @@ def parse_version(ver, pre=False):
     return Version(major, minor, micro, release, pre, post, dev)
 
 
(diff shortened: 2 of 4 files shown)


facelessuser:
@gir-bot lgtm

PR #2105, opened 2023-07-10
Combine parent header and content tab slug via new option

### LINKED ISSUE #2098: Slugs that combine headers with anchors
### Problem

When using Mkdocs Material, it generates slugs for links being clicked based on the text of the link. In particular, I'm using a feature called [Content Tab Anchor Links](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#anchor-links), where tabs are rendered on a page, and clicking a tab will update the URL based on the tab's text.

image

When clicking on the tabs from content tabs that have the same names, I get links like these:

- example.com/#about
- example.com/#value
- example.com/#about_1
- example.com/#value_1

What I'd really like is to have the headers factor into the slug:

- example.com/#log_file-about
- example.com/#log_file-value
- example.com/#log_level-about
- example.com/#log_level-value

I opened a [discussion over at Mkdocs Material](https://github.com/squidfunk/mkdocs-material/discussions/5675), and it was mentioned to see if this was possible or something that could be pursued over here.

Is there any way to achieve this today? I was trying to figure out if I could write my own python function, but I couldn't easily determine that nor could I determine if there was a way to get the header(s) preceding a content tab.

### Benefits

Allows having even more readable slugs by using contextual positioning to fully represent where you are in a page. I've seen this done with HTML output from tools like vimwiki where they keep track of the depth that elements are in within a page, and combine all of the nested headers into part of the anchor.

### Solution Idea

I'm unaware of how Python markdown parsing works in terms of a solid structure. If it forms any kind of tree or association between headers and other content, then it would be trivial to reverse up the stack to see what each header's text was and combine that.

vimiwki handles it by scanning a document from top to bottom and maintaining a stateful track of the current headers as a stack, pushing onto the stack when we see a smaller header, and popping off the stack when we encounter a larger header. When a link is encountered, the current header tracking is used as a prefix to the anchor itself.

```markdown
# One

## Two

some anchor text

## Three

### Four

some anchor text
```

Yields links `example.com/#one-two-some-anchor-text` and `example.com/#one-three-four-some-anchor-text`.

facelessuser:
Resolves #2098

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 10.1
+
+-   **NEW**: Add new `combine_header_slug` option in legacy Tabbed extension and new Block Tab extension that will
+    prefix a content tab's slug with the parent header's slug. This allows for content tab slugs that are scoped to the
+    header they are under.
+
 ## 10.0.1
 
 - **FIX**: Regression related to snippets nested deeply under specified base path.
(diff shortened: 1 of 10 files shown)


facelessuser:
@gir-bot lgtm

PR #2121, opened 2023-07-22
Add highlight lexer cache to improve performance

kamilkrzyskow:
Hi 👋,
adding this caching method to the script decreased `mkdocs serve` time from ~33* seconds down to ~7 seconds.
My rationale is that the lexer for a given language won't change and shouldn't change during runtime. If someone injects pygment lexers during runtime to change the highlight colour, they could also clear the cache 🤷 

🙋 Please let me know if this is something I should report upstream to Pygments and make them use `functools.lru_cache` for the `get_lexer_by_name` function, but I don't think adding caching to this script is wrong 🤔.

\* - the 33 seconds time is from my current shabby "setup". Old laptop with 2 cores and 2 threads, so this kind of exaggerates the performance gap, but still it's x5 performance gain ✌️  

---

At first I thought, our custom lexer is at fault (not optimized etc.), but removing the lexer module from the site-packages resulted in about the same time (~30 seconds). Setting the `use_pygments` param to `false` resulted in a runtime of ~4 seconds.
After I added the cache, the removal of the lexer module resulted in a runtime of ~5 seconds, so +2 seconds to process all of the code blocks via our custom lexer seems reasonable to me, considering my setup and the amount of documentation pages.

I tested it with our Gothic Modding Community documentation page. [This commit specifically](https://github.com/Gothic-Modding-Community/gmc/tree/1100c33ba44135b2a0210a4cc98010d25225af3f)
Together with a cut down minified `mkdocs.yml`:

<details>

```yml
site_name: Python Extensions Performance Test

watch:
  - overrides

theme:
  name: material
  custom_dir: overrides
  favicon: assets/images/gmc_logo.png
  logo: assets/images/gmc_logo.png

extra_css:
  - assets/stylesheets/extra.css

extra_javascript:
  - assets/javascripts/extra.js

markdown_extensions:
  - admonition
  - pymdownx.highlight:
      use_pygments: true
      linenums: true
      anchor_linenums: true
  - pymdownx.inlinehilite
  - attr_list
  - pymdownx.superfences:
      custom_fences:
        - name: mermaid
          class: mermaid
          format: !!python/name:pymdownx.superfences.fence_code_format
```

</details>

--- 

I also considered adding some sort of caching to the `inlinehilite`, because without testing/profiling, as I understand, it creates a new copy instance of the used highlighter class, each time it highlights inline code. The same rationale applies, the highlighter for a given language shouldn't change during runtime, so there is no need to search for it every time.

I ultimately didn't look into it further as commenting out `pymdownx.inlinehilite` from the `mkdocs.yml` file resulted in similar performance, it decreased the time maybe by ~0.5 seconds, therefore I don't think that adding caching would be as much beneficial 🤷

### DIFF
--- a/pymdownx/highlight.py
+++ b/pymdownx/highlight.py
@@ -39,6 +39,7 @@ except ImportError:  # pragma: no cover
     pygments = False
     p_ver = (0, 0)
 
+LEXER_CLASS_CACHE = {}
 RE_PYG_CODE = re.compile(r'^<(div)(\s*class="(.*?)")?\s*>')
 CODE_WRAP = '<pre{}><code{}{}{}>{}</code></pre>'
 CODE_WRAP_ON_PRE = '<pre{}{}{}><code>{}</code></pre>'
@@ -281,6 +282,9 @@ class Highlight(object):
 
         if language:
             language, lexer_options = self.get_extended_language(language)
+            cached_lexer_class = LEXER_CLASS_CACHE.get(language)
+            if cached_lexer_class:
+                return cached_lexer_class(**lexer_options), name
         else:
             lexer_options = {}
 
@@ -297,9 +301,15 @@ class Highlight(object):
                     name = lexer.aliases[0]
                 except Exception:  # pragma: no cover
                     pass
+        else:
+            LEXER_CLASS_CACHE[language] = lexer.__class__
+        
         if lexer is None:
             lexer = get_lexer_by_name('text')
             name = lexer.aliases[0]
+            if language:
+                LEXER_CLASS_CACHE[language] = lexer.__class__
+
         return lexer, name
 
     def escape(self, txt):


facelessuser:
We would have to look very close at this. I already see one issue. `language, lexer_options = self.get_extended_language(language)`. What you are proposing keeps a global cache, but if we have a script that spawns to different Markdown tasks, and both have different configs, the extended language could be different, but the cache will cause them to be the same. There could be other dynamic stuff, but I'm not sure.

kamilkrzyskow:
I run the tox test now and it turned out that lexer guessing tests are failing 🤔, so perhaps I shouldn't save the cache when it guesses the language 😮‍💨. I shall investigate.

facelessuser:
Oh, saving guess would be a no go.

kamilkrzyskow:
I missed the `language, lexer_options = self.get_extended_language(language)` part 👀 💦 
Between switching settings and profiling the speed it just slipped my mind that the language param of the function is later overwritten..., and I admit I probably don't understand what kind of edge-case there is when it comes to the extended language.

I moved things around to lookup cache after the `language` is overwritten, and cache based on the `language` used in the `get_lexer_by_name` function call, and I cache only after a successful function result.

Running tox with Python 3.11 locally results in a happy face 😄

kamilkrzyskow:
I did some digging and it turned out that Pygments does indeed have some sort of cache in place https://github.com/pygments/pygments/blob/master/pygments/lexers/__init__.py#L126
And they store classes in the cache to call them later with different lexer options.
I don't know why it doesn't work properly on their end (as in adding the cache here adds performance), but I changed my approach to store the class and the performance gain seems to be somewhat the same.

I think that is my final edit ✌️

facelessuser:
I will have to go over this to make sure there are no adverse effects. Additionally, to accept something like this, I will also have to ask you to provide a script to verify the claim of improved performance as well.

kamilkrzyskow:
OK, I will look into how MkDocs processes the Markdown files and try to replicate the methodology of using the extensions. I will try to setup a script which simulates processing a thousand files with some generic C++ and Python code blocks. I believe this should be enough to showcase the performance difference, but let me know if you want something else.

facelessuser:
We are testing a lot of things by going through MkDocs, you should be able to demonstrate an performance increase using Python Markdown directly.

kamilkrzyskow:
So my initial hunch of the issue being related to my custom lexer turned out to be correct. Each time the `get_lexer_by_name` function is called it traverses the whole LEXERS dictionary and then if not found searches the plugin lexers... which means that in my case it always run the worst case scenario of traversing the whole 500+ lexer list and then finding the custom lexer at the end. Just my luck. https://github.com/pygments/pygments/blob/master/pygments/lexers/__init__.py#L122-L130

I made this script [performance_test.zip](https://github.com/facelessuser/pymdown-extensions/files/12137225/performance_test.zip) 
> It has markdown included, so the Github formatting breaks it

Running it on the e6474b38703b45e3dc431d3c1a0b1f24d80ee7fa commit results in 
```
Running bash_lexer x 100 times
-> Elapsed time 0.2318s
Running python_lexer x 100 times
-> Elapsed time 0.4003s
Running custom_lexer x 100 times
-> Elapsed time 2.5689s
```

and running it on the PR HEAD results in 
```
Running bash_lexer x 100 times
-> Elapsed time 0.2337s
Running python_lexer x 100 times
-> Elapsed time 0.3827s
Running custom_lexer x 100 times
-> Elapsed time 0.3471s
```

The `bash` lexer is ~40th in order and the `python` lexer is ~378th in order, therefore there is this difference in speed.
However, I don't understand why the added cache, doesn't seem to affect the Bash and Python times at all, it only affects the worst case scenario which is accessing custom plugin lexers (no matter if they exist or not) 🤔

facelessuser:
Are you not registering your plugin as a proper package?

kamilkrzyskow:
I'm not sure what do you mean by "registering as proper package". I copied the project structure of another custom lexer and it matches what is said in the documentation https://pygments.org/docs/plugins/
This is my setup.py with an appropriate entry point https://github.com/kamilkrzyskow/gothic-lexer/blob/main/setup.py#L17-L20
Pygments is able to find the lexer, however it seems that the lookup speed of custom plugin lexers isn't ideal.
I'll keep digging as there is an issue somewhere and maybe my proposed cache isn't the best way to solve it.

facelessuser:
> This is my setup.py with an appropriate entry point https://github.com/kamilkrzyskow/gothic-lexer/blob/main/setup.py#L17-L20

Okay, just making sure. So custom lexers aren't cached and other lexers are? It was mentioned earlier that Pygments is supposed to be caching this. This makes me question if we are trying to patch this in the wrong place.

kamilkrzyskow:
I simplified the performance test to only have empty code blocks as the issue lies in the lookup and not the lexing itself. [performance_test_updated.zip](https://github.com/facelessuser/pymdown-extensions/files/12137383/performance_test.zip)

I also added some profiling to Pygments itself in the `lexers/__init__.py`: 
<details>

```py
import time

TIME_ACC = {}

def performance_decorator(func):
    def wrapper(*args, **kwargs):
        #if args[0] == "text":
        #    return
        print(f"    Running {func.__name__}")
        start = time.perf_counter()
        func(*args, **kwargs)
        ct = time.perf_counter() - start
        ts = TIME_ACC.get(args[0], 0) + ct
        TIME_ACC[args[0]] = ts
        print(f"    -> Current time for {args[0]} = {ct}s | Time sum = {ts}s")        
    
    return wrapper

@performance_decorator
def get_lexer_by_name(_alias, **options):
```

</details>

which results in this log:
<details>

```
Running bash_lexer x 10 times
    Running get_lexer_by_name
    -> Current time for shell = 0.016942699992796406s | Time sum = 0.016942699992796406s
    Running get_lexer_by_name
    -> Current time for text = 0.0054835000046296045s | Time sum = 0.0054835000046296045s
    Running get_lexer_by_name
    -> Current time for shell = 4.329999501351267e-05s | Time sum = 0.01698599998780992s
    Running get_lexer_by_name
    -> Current time for text = 0.0001429999974789098s | Time sum = 0.005626500002108514s
    Running get_lexer_by_name
    -> Current time for shell = 4.299999272916466e-05s | Time sum = 0.017028999980539083s
    Running get_lexer_by_name
    -> Current time for text = 0.0001387999946018681s | Time sum = 0.005765299996710382s
    Running get_lexer_by_name
    -> Current time for shell = 3.7299992982298136e-05s | Time sum = 0.01706629997352138s
    Running get_lexer_by_name
    -> Current time for text = 0.00013949999993201345s | Time sum = 0.005904799996642396s
    Running get_lexer_by_name
    -> Current time for shell = 3.619999915827066e-05s | Time sum = 0.017102499972679652s
    Running get_lexer_by_name
    -> Current time for text = 0.00014209999062586576s | Time sum = 0.006046899987268262s
    Running get_lexer_by_name
    -> Current time for shell = 3.670000296551734e-05s | Time sum = 0.01713919997564517s
    Running get_lexer_by_name
    -> Current time for text = 0.00013770000077784061s | Time sum = 0.006184599988046102s
    Running get_lexer_by_name
    -> Current time for shell = 3.599999763537198e-05s | Time sum = 0.01717519997328054s
    Running get_lexer_by_name
    -> Current time for text = 0.0001323000033153221s | Time sum = 0.006316899991361424s
    Running get_lexer_by_name
    -> Current time for shell = 4.610000178217888e-05s | Time sum = 0.01722129997506272s
    Running get_lexer_by_name
    -> Current time for text = 0.0001571000029798597s | Time sum = 0.006473999994341284s
    Running get_lexer_by_name
    -> Current time for shell = 8.099999104160815e-05s | Time sum = 0.01730229996610433s
    Running get_lexer_by_name
    -> Current time for text = 0.00033629999961704016s | Time sum = 0.006810299993958324s
    Running get_lexer_by_name
    -> Current time for shell = 4.490000719670206e-05s | Time sum = 0.01734719997330103s
    Running get_lexer_by_name
    -> Current time for text = 0.0001505999971413985s | Time sum = 0.006960899991099723s
-> Elapsed time 0.0381s
Running python_lexer x 10 times
    Running get_lexer_by_name
    -> Current time for py = 0.06926090001070406s | Time sum = 0.06926090001070406s
    Running get_lexer_by_name
    -> Current time for text = 0.00016169999435078353s | Time sum = 0.007122599985450506s
    Running get_lexer_by_name
    -> Current time for py = 0.0001311999949393794s | Time sum = 0.06939210000564344s
    Running get_lexer_by_name
    -> Current time for text = 9.420000424142927e-05s | Time sum = 0.0072167999896919355s
    Running get_lexer_by_name
    -> Current time for py = 0.0001230000052601099s | Time sum = 0.06951510001090355s
    Running get_lexer_by_name
    -> Current time for text = 9.159999899566174e-05s | Time sum = 0.007308399988687597s
    Running get_lexer_by_name
    -> Current time for py = 0.00012209999840706587s | Time sum = 0.06963720000931062s
    Running get_lexer_by_name
    -> Current time for text = 9.560000034980476e-05s | Time sum = 0.007403999989037402s
    Running get_lexer_by_name
    -> Current time for py = 0.00012080000306013972s | Time sum = 0.06975800001237076s
    Running get_lexer_by_name
    -> Current time for text = 9.31999966269359e-05s | Time sum = 0.007497199985664338s
    Running get_lexer_by_name
    -> Current time for py = 0.0001213000068673864s | Time sum = 0.06987930001923814s
    Running get_lexer_by_name
    -> Current time for text = 0.00011150000500492752s | Time sum = 0.007608699990669265s
    Running get_lexer_by_name
    -> Current time for py = 0.00012139999307692051s | Time sum = 0.07000070001231506s
    Running get_lexer_by_name
    -> Current time for text = 9.460000728722662e-05s | Time sum = 0.007703299997956492s
    Running get_lexer_by_name
    -> Current time for py = 0.00017250000382773578s | Time sum = 0.0701732000161428s
    Running get_lexer_by_name
    -> Current time for text = 0.0006011999939801171s | Time sum = 0.00830449999193661s
    Running get_lexer_by_name
    -> Current time for py = 0.00014150000060908496s | Time sum = 0.07031470001675189s
    Running get_lexer_by_name
    -> Current time for text = 9.780000254977494e-05s | Time sum = 0.008402299994486384s
    Running get_lexer_by_name
    -> Current time for py = 0.00012719999358523637s | Time sum = 0.07044190001033712s
    Running get_lexer_by_name
    -> Current time for text = 9.420000424142927e-05s | Time sum = 0.008496499998727813s
-> Elapsed time 0.0812s
Running custom_lexer x 10 times
    Running get_lexer_by_name
    -> Current time for dae = 0.03511510000680573s | Time sum = 0.03511510000680573s
    Running get_lexer_by_name
    -> Current time for text = 0.0001618999958736822s | Time sum = 0.008658399994601496s
    Running get_lexer_by_name
    -> Current time for dae = 0.033102899993536994s | Time sum = 0.06821800000034273s
    Running get_lexer_by_name
    -> Current time for text = 0.00023380000493489206s | Time sum = 0.008892199999536388s
    Running get_lexer_by_name
    -> Current time for dae = 0.02337660000193864s | Time sum = 0.09159460000228137s
    Running get_lexer_by_name
    -> Current time for text = 0.00016589999722782522s | Time sum = 0.009058099996764213s
    Running get_lexer_by_name
    -> Current time for dae = 0.028981900002690963s | Time sum = 0.12057650000497233s
    Running get_lexer_by_name
    -> Current time for text = 0.00017499999376013875s | Time sum = 0.009233099990524352s
    Running get_lexer_by_name
    -> Current time for dae = 0.030043500009924173s | Time sum = 0.1506200000148965s
    Running get_lexer_by_name
    -> Current time for text = 0.00017940001271199435s | Time sum = 0.009412500003236346s
    Running get_lexer_by_name
    -> Current time for dae = 0.031268399994587526s | Time sum = 0.18188840000948403s
    Running get_lexer_by_name
    -> Current time for text = 0.00019559999054763466s | Time sum = 0.00960809999378398s
    Running get_lexer_by_name
    -> Current time for dae = 0.031182200007606298s | Time sum = 0.21307060001709033s
    Running get_lexer_by_name
    -> Current time for text = 0.00018030000501312315s | Time sum = 0.009788399998797104s
    Running get_lexer_by_name
    -> Current time for dae = 0.02898550000099931s | Time sum = 0.24205610001808964s
    Running get_lexer_by_name
    -> Current time for text = 0.00020780001068487763s | Time sum = 0.009996200009481981s
    Running get_lexer_by_name
    -> Current time for dae = 0.026904900005320087s | Time sum = 0.2689610000234097s
    Running get_lexer_by_name
    -> Current time for text = 0.0001838999887695536s | Time sum = 0.010180099998251535s
    Running get_lexer_by_name
    -> Current time for dae = 0.04005739999411162s | Time sum = 0.30901840001752134s
    Running get_lexer_by_name
    -> Current time for text = 0.0001795000134734437s | Time sum = 0.010359600011724979s
-> Elapsed time 0.3361s
```

</details>

Which revealed that for some reason the `text` lexer is run too even though I explicitly stated the languages in markdown. It also shows that the built-in lexers do indeed use some sort of cache (even though it's weird that bash and python have such different lookup times from cache) and that external lexers don't use the cache at all 😞.

Thanks for the quick responses and guiding me toward the right places ✌️

facelessuser:
Interesting, well I'm glad I could help 🙂

PR #2206, opened 2023-10-14
Support PathLike objects in base_path

oprypin:
This enables this usage in MkDocs (note that specifying as a list already works):

```yml
  - pymdownx.snippets:
      base_path: !relative $docs_dir
```

Currently you get an error:
```
               self.base_path = [os.path.abspath(b) for b in base]
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
           TypeError: 'DocsDirPlaceholder' object is not iterable
```


facelessuser:
We would need tests to confirm that things are working with PathLike objects.

facelessuser:
Can you enable `check_paths` to make sure that portion works as well?

facelessuser:
Hmm, it may not be necessary to enable `check_paths`. I looked over the code again and it seems we test the file's existence whether it is enabled or not, `check_paths` just raises an error if the file is not found. Since the test demonstrates the inserted content is there, we can assume it is checking and finding things as needed.

oprypin:
So does anything need to be changed? I understood this as probably not?

facelessuser:
No, I think it is fine, thanks for checking back in.

facelessuser:
@gir-bot lgtm

PR #2241, opened 2023-11-19
Add further customization ability to admonitions and details

facelessuser:
- Allow specifying a specific class that is not the same as the custom block name.
- Allow specifying a default title that is not derived from the class.

Also remove some dead code that was used during development of the block feature.


facelessuser:
@gir-bot lgtm

PR #2245, opened 2023-11-21
Keys should not parse base64 encoded URLs

### LINKED ISSUE #2243: pymdownx.keys may break base64 encoded images
### Description

With a base64 encoded image, given the fact that a base64 encoded string can contain something like `++a++`, the pymdownx.keys would convert it to `A` which will corrupt the content.

pymdownx.keys should perhaps not process strings within a link since the link URL should always be left as-is.

### Minimal Reproduction

1. Enable pymdownx.keys in mkdocs.yml.
2. Create a markdown file with content like this: ``
3. Build and browse to the site, and the resulting image is ``

### Version(s) & System Info

- Operating System: Windows 10
- Python Version: 3.12
- Package Version: latest

facelessuser:
Fixes #2243

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 10.5
+
+- **FIX**: Keys: Ensure that Keys does not parse base64 encoded URLs.
+
 ## 10.4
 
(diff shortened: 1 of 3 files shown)


facelessuser:
@gir-bot lgtm

PR #2413, opened 2024-07-14
fix: allow downloading when content-length header is missing

### LINKED ISSUE #2404: Download snippet + server reply without content-length
### Description

Hi,

When I use:
```
        --8<-- "http://localhost:3001/files/maartenbreddels/test/app.py"
```

But my server replies with:
```
$ curl -I http://localhost:3001/files/maartenbreddels/test/app.py
HTTP/1.1 200 OK
cache-control: public, max-age=0, must-revalidate
server: Vercel
x-vercel-id: dev1::qcjlo-1720123058986-b8c5fdc87ba8
x-vercel-cache: MISS
vary: RSC, Next-Router-State-Tree, Next-Router-Prefetch, Accept-Encoding
content-type: text/plain
date: Thu, 04 Jul 2024 19:57:39 GMT
keep-alive: timeout=5
connection: close
```

(i.e. no content-length - which is also possible with `transfer-encoding: chunked`)

This extension raises the error: 
```
                raise ValueError("Missing content-length header")
```

A possible solution could be not to do the check when `url_max_size==0`.

Let me know what you think.

Regards,

Maarten

### Minimal Reproduction

I cannot find a public server that does this, so difficult to give you reproducible steps now

### Version(s) & System Info

markdown-it-py                            3.0.0

maartenbreddels:
This can happen for valid http responses, such as when the server is using chunked encoding.

Fixes #2404.


maartenbreddels:
This makes the following work

```py
--8<-- "https://py.cafe/files/maartenbreddels/altair-car-performance-comparison/app.py"
```

Which gives these headers:
```
$ curl -I https://py.cafe/files/maartenbreddels/altair-car-performance-comparison/app.py                                                   fix_download_without_content_length_header ◼
HTTP/2 200 
age: 0
cache-control: public, max-age=0, must-revalidate
content-type: text/plain
date: Sun, 14 Jul 2024 19:45:46 GMT
server: Vercel
strict-transport-security: max-age=63072000
vary: RSC, Next-Router-State-Tree, Next-Router-Prefetch
x-matched-path: /files/[username]/[projectname]/[...path]
x-vercel-cache: MISS
x-vercel-execution-region: iad1
x-vercel-id: fra1::iad1::zhtpq-1720986346756-476b90c31ef2
```

PS: this is used to render the sourcecode for the hosted app at https://py.cafe/maartenbreddels/altair-car-performance-comparison

facelessuser:
Seems like a pretty reasonable approach.

We can ignore any lint errors outside of your changes (looks like some new checks are now throwing errors in old code that used to pass).

Tests will have to be updated to pass for new expectations as before we used to always error out for missing content length. I imagine we will likely need at least one new test to cover this new use case as well. If this seems like a little beyond what you are comfortable doing (as it will probably involve doing some mocking and such), I can probably take what you have and finish, but feel free to update if you are comfortable in this area. Either way, I think this is a decent way to move forward on this.

maartenbreddels:
Hi,

no problem with adding the test.
I managed to get the unit tests working. Let me know if you prefer to take over and change it to your liking, if so, feel free to force push on my branch.

Hope this is to your liking.

Regards,

Maarten

pymdownx/snippets.py
```diff
@@ -199,10 +199,14 @@ def download(self, url):
 
             # We provide some basic protection against absurdly large files.
             # 32MB is chosen as an arbitrary upper limit. This can be raised if desired.
-            length = response.headers.get("content-length")
-            if length is None:
-                raise ValueError("Missing content-length header")
-            content_length = int(length)
+            content = None
+            if "content-length" not in response.headers:
+                # we have to read to know if we went over the max, but never more than url_max_size
```
facelessuser, line 204: Please wrap `url_max_size` in backticks (`` ` ``) so that the spellchecker will ignore them.

tests/test_extensions/test_snippets.py
```diff
@@ -892,17 +892,13 @@ def test_url_nested_file(self, mock_urlopen):
     def test_url_lines(self, mock_urlopen):
         """Test specifying specific lines in a URL."""
 
-        content = []
-        length = 0
-        with open('tests/test_extensions/_snippets/lines.txt', 'rb') as f:
-            for l in f:
-                length += len(l)
-                content.append(l)
+        content = open('tests/test_extensions/_snippets/lines.txt', 'rb').read()
```
facelessuser, line 895: While I'm indifferent as to whether `open` or `with open` is used, if `with` is not used, please explicitly close the file handle. `with` will take care of this for you. I understand that Cpython (specifically) will likely garbage collect this due to it not having a reference and going out of scope, but that is not the case with all Python implementations and I would like it to be explicitly closed, whether with `with` or without it.

facelessuser:
Overall it looks good, but I'd made a few comments.

maartenbreddels:
Good comments 👍 . I implemented them as two separate commits to help review. Feel free to squash merge it when ready.

facelessuser:
@maartenbreddels Thanks for your work on this!

@gir-bot lgtm

maartenbreddels:
Not to put pressure, but what is your release cadence?

I want to advertise to people to use pycafe + mkdocs and this plugin to have live code examples in their docs, which depends on this fix.

Could you let me know here or in #2404 when this is released, then I can let people know this is possible.

Thanks!

facelessuser:
I imagine it will be sometime soon. I can't promise I'll remember to comment. The best way to know when a release is made is to watch the repo, and if all you care about is releases, configure it to only watch releases. You can disable the watch once the release in question is made.

maartenbreddels:
Really good idea, will do that! Thanks a ton, also, for these great extensions, and from what I've seen a very nice and high-quality codebase. Cheers

PR #2449, opened 2024-09-07
fix: allow `SmartAsteriskProcessor` support inline bold and italic which are without spaces

andychucs:
Markdown allows the use of bold and italic without spaces. This expression causes a problem when it is used in without space case.
截屏2024-09-08 01 27 02

Some languages never require whitespace between words or phrases, like Chinese and Japanese.  

For example:

Correct case:  
截屏2024-09-08 01 19 53
截屏2024-09-08 01 28 39

Actual result in some docs:
截屏2024-09-08 01 25 52
截屏2024-09-08 01 26 15

### DIFF
--- a/pymdownx/betterem.py
+++ b/pymdownx/betterem.py
@@ -86,21 +86,21 @@ SMART_UNDER_EM2 = r'(?<![\w_])(_)(?![_\s])(.+?)(?<![_\s])(_)(?![_\w])'
 
 # Smart rules for when "smart asterisk" is enabled
 # SMART: ***strong,em***
-SMART_STAR_STRONG_EM = r'(?:(?<=_)|(?<![\w\*]))(\*{3})(?![\s\*])%s(?<!\s)\1(?:(?=_)|(?![\w\*]))' % SMART_STAR_CONTENT
+SMART_STAR_STRONG_EM = r'(?:(?<=\w)|(?<![\w\*]))(\*{3})(?![\s\*])%s(?<!\s)\1(?:(?=\w)|(?![\w\*]))' % SMART_STAR_CONTENT
 # ***strong,em* strong**
 SMART_STAR_STRONG_EM2 = \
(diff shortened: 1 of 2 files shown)


facelessuser:
BetterEm is already working as expected. Smart mode specifically does not allow mid-word bolding or italics. Smart mode is usually best for things like underscores which is why we only enable it on underscore by default. So if you don't want that behavior on asterisks, use the defaults which specifically do not enable smart mode for asterisks. Making smart mode behave other than what it is designed to do is not going to be allowed.

This behavior is enabled for underscore only by default and follows the behavior of GitHub and others.

---

```
smart_behavior_example

smart _behavior_ example

dumb*behavior*example

dumb *behavior* example
```

smart_behavior_example

smart _behavior_ example

dumb*behavior*example

dumb *behavior* example

---

So if what you want is like what you see above, don't enable smart mode on asterisks or you will get the underscore behavior on asterisks.

andychucs:
So if I want the default behavior like GitHub, then it should be used like following config
```
  pymdownx.betterem:
    smart_enable: none
```
```
  pymdownx.betterem:
    smart_enable: underscore
```
It still has some different behavior without using pymdownx.betterem? Should I just not use it in Chinese or Japanese docs?

-------

I don't found this example in BetterEm's documentation. Can I submit a pr to show this case in docstring and documentation, test case?

facelessuser:
@andychucs My point is that you don't need to set any of the options if you want only underscore to use the smart behavior, that's the default. See the documentation: https://facelessuser.github.io/pymdown-extensions/extensions/betterem/#options. Also, check out the example below:

```py
import markdown

MD = """
test**foo**test

test **foo** test

test**foo!**test

test*foo*test

test *foo* test

test*foo!*test

"""

print(markdown.markdown(MD, extensions=['pymdownx.betterem']))
```

```html
<p>test<strong>foo</strong>test</p>
<p>test <strong>foo</strong> test</p>
<p>test<strong>foo!</strong>test</p>
<p>test<em>foo</em>test</p>
<p>test <em>foo</em> test</p>
<p>test<em>foo!</em>test</p>
```

facelessuser:
While it is stated in the options, I can probably make it a bit more clear exactly what the defaults are as I do seem to leave it a bit ambiguous when I describe the plugin generally.

andychucs:
@facelessuser I get it. Thanks.

PR #2450, opened 2024-09-08
MagicLink: change processing to use x instead of twitter

facelessuser:
twitter is still available but deprecated and will be removed in the future.


facelessuser:
@gir-bot lgtm

PR #2464, opened 2024-09-22
fancylist inject class remove semicolon

joapuiib:
Removes the semicolon at the end of the injected class in `fancylists`.

### DIFF
--- a/pymdownx/fancylists.py
+++ b/pymdownx/fancylists.py
@@ -215,7 +215,7 @@ class FancyOListProcessor(BlockProcessor):
                     if self.inject_style:
                         attrib['style'] = f"list-style-type: {OL_STYLE[attrib['type']]};"
                     if self.inject_class:
-                        attrib['class'] = f"fancylists-{OL_STYLE[attrib['type']]};"
+                        attrib['class'] = f"fancylists-{OL_STYLE[attrib['type']]}"
                     lst = etree.SubElement(
                         parent,
                         self.TAG,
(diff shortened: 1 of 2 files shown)


joapuiib:
Oh, I assumed it was a typo. I've never seen semicolons in HTML classes.

Isn't more common not to use them?

facelessuser:
> Oh, I assumed it was a typo.

No, not a typo, but I honestly don't have strong feelings either way. It seemed like it was stated as a problem, but I don't think it really matters.

> Isn't more common not to use them?

It wouldn't surprise me if one was somewhat more popular than others. I know people that some have definite preference for or against them at the end.

I don't know what the split on preference is. I think at some point it used to be required, but I think these days most browsers are more forgiving. If you go online, you'll often see it taught to always add the `;` probably because it used to be needed, but yeah, I know there are people who don't add them and I've probably left them off at times myself. This is probably the most thought I've put into the significance of trailing commas in inline styles 🙃.

I guess my point is I don't really care it's there or not, but I guess I'm not against removing it either 🤷🏻.

joapuiib:
> This is probably the most thought I've put into the significance of trailing semicolons in inline styles 🙃.

I'm talking about setting the element `class` attribute, not `style`.

With the current implementation, CSS selector should be `.fancylists-lower-alpha\;` instead of `.fancylists-lower-alpha`.

Example:

```css
.fancylists-lower-alpha\; li::marker {
    content: counter(list-item, lower-alpha) ") ";
}
.fancylists-upper-alpha\; li::marker {
    content: counter(list-item, upper-alpha) ") ";
}
```

facelessuser:
Oh, a semicolon in the class?! Oh, that is definitely a typo.

facelessuser:
I completely missed that

facelessuser:
@gir-bot lgtm

joapuiib:
Sorry for the misunderstanding. I should have written a better PR text.

Thanks!

facelessuser:
I should have looked over the review closer 😅

facelessuser:
Released: https://github.com/facelessuser/pymdown-extensions/releases/tag/10.10.1

PR #2473, opened 2024-09-29
✏️ Fix typo in `extensions/blocks/api.md`

tiangolo:
✏️ Fix typo in `extensions/blocks/api.md`


facelessuser:
Not really a typo, but a stylistic change. While the edit is technically more concise, removing unnecessary phrasing, there is nothing wrong with the original phrasing.

facelessuser:
So, apparently, I missed it when I first looked at this PR that you were also addressing a doubled `the`. That is an actual type and a good catch.  I'm unfortunately prone to duplicating words. I decided to do a search for duplicated words in my documents and found quite a few and addressed them here: 3f6fbbb633aefb638c1c3eae1bb868824d80360d.

As for the `that`,  I actually prefer how it is read with it in place, so I'll be leaving it in as it is a stylistic preference.  Thank you for pointing out typo though.

As I've addressed related typos in the aforementioned commit, I will be closing this PR.

PR #2494, opened 2024-10-09
block caption extension: basic and autoid tests

joapuiib:
I've added tests for using the caption block extension with no configuration and `autoid` enabled.

The last test fails, when two captions with nested captions are present.


facelessuser:
I'll pull these in and adjust as necessary.

PR #2533, opened 2024-11-19
Fix broken link to options in block-caption plugin page

myl7:
This PR replaces `#options` with `#global-options` in that page.

An alternative method is to add `#options` anchor to the `Global Options` title.
This method makes the page become similar to other pages, and should be better.


facelessuser:
Good catch!

facelessuser:
@gir-bot lgtm

PR #2552, opened 2024-12-22
Allow negative indexes

### LINKED ISSUE #2517: Allow indexing from end
### Description

In the current state, when including specific line numbers, we can target specific lines.

1. To set starting line number: `myfile:3`
    * This will include the entire file STARTING at line number 3.
2. For lines up to a specific line: `myfile::10`
    * This will include lines 1 - 10
3. For a range of lines: `myfile:4:12`
    * This will include lines 4 - 12.
4. To include a SPECIFIC line: `myfile:3:3`
    * This will only include line number 3
    
    
BUT, what if we wanted to include the last 10 lines of the file. There currently is not a way to index from the end of the file.

### Benefits

Easy feature, which adds more utility, for potentially niche use-cases.

### Solution Idea

Add the ability to do negative indexing:

Given, a file that is 100 lines long-

1. To set starting line number: `myfile:-3`
    * This will include lines 97-100
 2. For lines up to a specific line: `myfile::-51`
    * This will include lines 1 - 49
3. For a range of lines: `myfile:4:-25`
    * This will include lines 4 - 75.
4. To include a SPECIFIC line: `myfile:-0:-0`
    * This will only include the very last line.

facelessuser:
Resolves #2517

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,8 +1,11 @@
 # Changelog
 
-## 10.12.1
+## 10.13
 
 -   **NEW**: Snippets: Allow multiple line numbers or line number blocks separated by `,`.
+-   **NEW**: Snippets: Allow using a negative index for number start indexes and end indexes. Negative indexes are
+    converted to positive indexes based on the number of lines in the snippet.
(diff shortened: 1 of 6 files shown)


facelessuser:
@gir-bot lgtm

PR #2560, opened 2025-01-09
Fix missing dict key in arithmatex docs

wilcoxjay:


facelessuser:
Good catch!

@gir-bot lgtm

PR #2634, opened 2025-03-25
add type annotations to block.py

### LINKED ISSUE #2633: Add type hints to project
### Description

I'm working on an extension that depends on this package, and would love to have type hints. Would a PR adding those be helpful?

### Benefits

The usual LSP support catching errors without running the code.

### Solution Idea

I'd be happy to open a PR.

superlopuh:
@facelessuser I added some type annotations to the functions I use in my extension, `uv run python -m mypy pymdownx/blocks` is happy with this, how can I check whether the testing command that I added is added correctly?

Resolves #2633


facelessuser:
While I'm okay with partial types, I think there is an expectation that if a file is touched to add types that the entire file would have types added.

Tests should be run via: `python3 -m tox -e py313` etc. You can also just run `mypy` via python3 -m mypy. You can certainly do this in a virtual environment as well.

I linked an example earlier, and like the example I linked, I would expect to see an explicit mypy configuration: https://github.com/facelessuser/wcmatch/blob/main/pyproject.toml#L62. Showing error codes is a must like in the linked example.

superlopuh:
Sounds good, will add this now.

pymdownx/blocks/block.py
```diff
@@ -130,13 +135,13 @@ def type_html_identifier(value):
...
 
-def _delimiter(string, split, string_type):
+def _delimiter(string: Any, split: str, string_type: Callable[[Any], str]) -> list[str]:
     """Split the string by the delimiter and then parse with the parser."""
 
     l = []
     # Ensure input is a string
-    string = type_string(string)
-    for s in string.split(split):
+    _string = type_string(string)
```
superlopuh, line 143: my lsp was happier with a new name for the type-restricted value, I'm not sure that it's strictly required by mypy. I think it makes things more explicit, but I'm happy to revert.

pymdownx/blocks/block.py
```diff
@@ -196,6 +201,17 @@ def type_html_attribute_dict(value):
 # Ensure class(es) or fail
 type_html_classes = type_string_delimiter(' ', type_html_identifier)
 
+from typing import Protocol
+
+
+class _BlocksProcessor(Protocol):
```
superlopuh, line 207: I introduced a new protocol to avoid moving around imports between the files, this has all the methods of BlocksProcessor that are used in this file.

pymdownx/blocks/block.py
```diff
@@ -205,10 +221,10 @@ class Block(metaclass=ABCMeta):
     NAME = ''
 
     # Instance arguments and options
-    ARGUMENT = False
-    OPTIONS = {}
+    ARGUMENT: bool | None = False
+    OPTIONS: dict[str, Any] = {}
```
superlopuh, line 225: not sure if this should be more restrictive

pymdownx/blocks/block.py
```diff
@@ -232,53 +248,55 @@ def __init__(self, length, tracker, block_mgr, config):
...
 
-    def dedent(self, text, length=None):
+    def dedent(self, text: str, length: int | None = None) -> str:
         """Dedent raw text."""
 
         if length is None:
             length = self.md.tab_length
 
-        min_length = float('inf')
+        min_length = length
```
superlopuh, line 280: this is the only "logic change", if you can call it that, avoiding the float in the mix between integer comparisons, and relying on the commutativity of `min` to start with the last value to `min` with.

pymdownx/blocks/block.py
facelessuser: It's not because of the rename, it's because the input is typed as any and then the function returns string to the same variable. Using a new variable ensures causes the new variable to be scooped to string only.

pymdownx/blocks/block.py
```diff
@@ -30,39 +35,39 @@ def _type_multi(value, types=None):
     raise ValueError(f"Type '{type(value)}' did not match any of the provided types")
 
 
-def type_multi(*args):
+def type_multi(*args: Callable[[Any], _T]) -> Callable[[Any], _T]:
```
facelessuser, line 38: Isn't `*args` an interable?

pymdownx/blocks/block.py
```diff
@@ -165,13 +170,13 @@ def type_string_in(accepted, insensitive=True):
...
 
-def type_string_delimiter(split, string_type=type_string):
+def type_string_delimiter(split: str, string_type: Callable[[Any], str] = type_string) -> Callable[[Any], list[str]]:
     """String delimiter function."""
 
     return functools.partial(_delimiter, split=split, string_type=string_type)
 
 
-def type_html_attribute_dict(value):
+def type_html_attribute_dict(value: dict[str, str | list[str]]) -> dict[str, str | list[str]]:
```
facelessuser, line 179: Many of these accept `Any` for `value`, even this one. They validate and raise an exception if not of the expected type. Technically, you could scope the input to the expected type and have the exception raise if not, or accept Any and ensure the return is as expected. Regardless, we need to be consistent in what we do. Due to the dynamic nature, it may be easiest to just always accept `Any` for value and ensure the return type is as expected.

pymdownx/blocks/block.py
superlopuh: The way they set up the type system is that for `*varargs` the type is of the element of the n-tuple

facelessuser: Yeah, you're right. Now I remember I've done this as well in the past.

pymdownx/blocks/block.py
superlopuh: This isn't the problem here. It's actually that there is a circular dependency between the two files, as BlocksProcessor processes blocks, and a Block holds a reference to a BlockProcessor. This is fine in practice as it's an implicit circular dependency, so the client will have imported both files before doing things with either class, but it poses an issue for typing.

superlopuh: Typing _BlocksProcessor won't solve the issue here, it would need a larger refactor, probably by creating a superclass of BlocksProcessor with the relevant functionality. I'm happy to do that if you would prefer it to the protocol.

facelessuser: This should fix your issues of import

```py
# types defined in function headers will be treated as strings during Runtime.
# This is needed until this is default in all supported Python versions.
from __future__ import annotations
...
# Import only during type checking, not needed during runtime
if TYPE_CHECKING:  # pragma: no cover
    from ..blocks import BlocksProcessor
...
    # Use as necessary in function headers. If you must use it outside of such
    # places, you can use comments to type `# type: BlockProcessor` and even cast
    # by using a literal string `cast('BlockProcessor', obj)`.
    def __init__(self, length, tracker, block_mgr: BlocksProcessor, config):
```

pymdownx/blocks/block.py
facelessuser: Just use `sys.maxsize` instead. You are correct though, a float really shouldn't be used here.

pymdownx/blocks/block.py
```diff
@@ -1,13 +1,19 @@
 """Block class."""
+from __future__ import annotations
+
```
facelessuser, line 3: Extra newline is unnecessary here.

pymdownx/blocks/block.py
facelessuser: Can you elaborate? There seems to be some concern here, but I would need a bit more reasoning as to why there is concern here.

pymdownx/blocks/block.py
superlopuh: Done

pymdownx/blocks/block.py
superlopuh: Done

pymdownx/blocks/block.py
superlopuh: Done

pymdownx/blocks/block.py
facelessuser: Yeah, so it is a dict with a string key, but currently it is defined with a list of of two items: a default, and a function to parse the current input.

Obviously, list is a bad choice if typing. Tuple would be better because then we could be more specific: `dict[str, tuple[Any, Callable[[Any], _T]]` (or something like that).

facelessuser: If we want to do something like this, we need to adjust existing extensions to use a Tuple and update documentation to showcase the tuple instead of lists as it currently does.

superlopuh: I would welcome that change but it feels like it belongs in a separate PR

facelessuser: I'm not sure I agree, the point of this PR is to provide good, accurate types for this submodule, and it should be a non-functional change.

superlopuh: Maybe I'm misunderstanding, what is the change that you propose?

facelessuser: Sorry, updated the last statement to specify the tuple should contain a default and a callback, not a string and callback.

superlopuh: I see, I would usually split this kind of behaviour change into a separate PR from a type annotation change. How would you feel about me opening that PR first, merging that, and then coming back to this one to add the types?

facelessuser: As the reviewer and the maintainer, I don't mind these changes being combined as these are related to getting this submodule properly typed, but if you want to do a separate PR, I won't stop you.

superlopuh: Fair enough, I'll add it to this one within the next few days.

superlopuh: I updated the type and the examples, please let me know if I've missed anything.

superlopuh:
Gentle ping :)

superlopuh:
When I run things locally, it succeeds:

```
~/Dev/f/pymdown-extensions sasha/type-hints-blocks !4 ?1 ❯ mypy                                                                                                            Py pymdown-extensions 13:28:51
Success: no issues found in 1 source file
```

Do you know how I can reproduce the errors locally to remove them?

facelessuser:
Sorry, been distracted.

So it looks like we may need to somehow explicitly ignore some of the imported submodules. I think mypy is trying to check the types of them simply because they are imported into the one we are interested in. I think maybe there is a code that can prevent this globally, or an ignore we can place online on the specific import causing the failures.

Also, we have one lint error.

facelessuser:
I'm sorry, I gave the wrong instructions, it's `python3 -m tox -e py313` (using the python number appropriate for you system). I tested locally and mypy is working fine...

facelessuser:
Never mind, I was on the wrong branch. It fails locally for me when run with tox

pymdownx/blocks/block.py
```diff
@@ -331,10 +345,11 @@ def on_validate(self, parent):
         return True
 
     @abstractmethod
-    def on_create(self, parent):
+    def on_create(self, parent: etree.Element) -> etree.Element:
         """Create the needed element and return it."""
+        ...
```
facelessuser, line 350: Remove `...`, it isn't needed when there is a docstring and is just causing test coverage issues.

pymdownx/blocks/block.py
```diff
@@ -1,9 +1,17 @@
...
 import functools
 import copy
 import re
+import sys
 from markdown import util as mutil
+import xml.etree.ElementTree as etree
+from typing import Any, Callable, TypeVar, TYPE_CHECKING
+from collections.abc import Iterable
+
+if TYPE_CHECKING:
```
facelessuser, line 13: Please add the comment `# pragma: no cover` on the if line to ignore code coverage here.

facelessuser:
This will solve the mypy issues:

```toml
[[tool.mypy.overrides]]
module = [
    "pymdownx.util",
    "pymdownx.__meta__",
    "pymdownx.superfences",
    "pymdownx.blocks"
]
ignore_errors = true
```

In general it is better to start at the bottom and work your way up with typing as you won't have complete good typing starting at the upper layers. That is why we have these problems, we started at a layer above all the other untyped code.

superlopuh:
Nice, thank you! It does seem like a part of the problem is with running mypy with tox not using the config in the pyproject.toml, but I didn't figure out how to link them up properly, your ignore list works for me.

facelessuser:
As far as I'm aware tox is using the config, but maybe I don't understand what you mean. Regardless, things seem to be working now. Please address the lint issue and I'll take a final pass.

facelessuser:
This is probably fine. I'll likely merge it as is and if I need to tweak something I will.

superlopuh:
Gentle ping :)

facelessuser:
I haven't forgotten, I've just been working on some other things. The plan is to include this in the next release. As far as this PR is concerned, it is ready. I'll likely merge once I'm done with the other things I'm working on.

superlopuh:
Great, thank you!

facelessuser:
@gir-bot lgtm

PR #2649, opened 2025-04-21
Provide option for more relaxed fenced code headers

### LINKED ISSUE #2544: Compatibility issues with Highlighting Lines in Code blocks
### Description

I have a paragraph in my markdown text like this, and when I run it, it parses incorrectly, causing the rendered page to be completely misaligned, here's the source file: [TestOne.md](https://github.com/user-attachments/files/18033885/TestOne.md).

````
```javascript [4]
function hello(name) {
    let phrase = `Hello, ${name}!`;

    debugger;

    say(phrase);
}
```
````

<mark>It would be rendered like this:</mark>

iShot_2024-12-06_14 32 12

---

<mark>This is the correct parsing in Typora：</mark>

iShot_2024-12-06_14 32 36

<br />The reason is that the [4] after the token is not recognized, and it treats the [4] as the beginning of the code, which should normally be the next line.

I know this is not the canonical writeup for this project, but it could be designed so that ignore unrecognized markup when it is encountered, thus not affecting the parsing and rendering of later text.This writeup renders fine in other editors, such as Typora, Github and so on.

This writeup is from other project [revealjs](https://revealjs.com/code/), I compared, this design is much simpler, just need to set the number, don't need to remember the property name, and it also supports step-by-step highlights.

- If you agree with this design, you can consider adding support in subsequent updates;
- If other settings are involved and it is troublesome to modify, then forget it, but it needs to be designed to ignore the markup and not affect the analysis and rendering of the following text;

Here are some write-ups for comparison and reference:

iShot_2024-12-05_12 40 08

### Minimal Reproduction

unnecessary

### Version(s) & System Info

unnecessary

facelessuser:
Resolves #2544

### DIFF
--- a/docs/src/markdown/about/changelog.md
+++ b/docs/src/markdown/about/changelog.md
@@ -1,7 +1,10 @@
 # Changelog
 
-## Unreleased
+## 10.15.0
 
+-   **NEW**: SuperFences: Add `relaxed_headers` option which can tolerate bad content in the fenced code header. When
+    enabled, code blocks with bad content in the header will likely still convert into code blocks, often respecting
+    the specified language.
 -   **FIX**: Blocks: Fix some corner cases of nested blocks with lists.
 -   **FIX**: Tab and Tabbed: Fix a case where tabs could fail if `combine_header_slug` was enabled and there was no
     header.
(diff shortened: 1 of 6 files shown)


facelessuser:
@gir-bot lgtm

Issue #2491, opened 2024-10-08
IDE Formating and blocks attributes


Leikt:
### Description

Hello, first thanks for the amazing work you've been doing on Pymdown !

There is a inconvenience when using an IDE to reformat the markdown documents.

When reformatting something like this :
```markdown
/// admonition | Sample warning
    type: warning
The content of my warning...
///
```

It becomes :
```markdown
/// admonition | Sample warning
type: warning
The content of my warning...
///
```

That is annoying because it breaks the block feature.

**Solution idea**

To fix this, it should be possible to use a non whitespace prefix. Per example:

```markdown
/// admonition | Sample warning
{type: warning, attrs: {style: 'font-size: xx-large'}}
The content of my warning...
///
```

or

```markdown
/// admonition | Sample warning
# type: warning
# attrs: {style: 'font-size: xx-large'}
The content of my warning...
///
```

### Minimal Reproduction

1. Create a new mkdocs-material project: https://squidfunk.github.io/mkdocs-material/getting-started/
2. Install Pymdown: https://facelessuser.github.io/pymdown-extensions/installation/
3. Enable the pymdown.block.admonition feature: https://facelessuser.github.io/pymdown-extensions/extensions/blocks/ (or add `pymdownx.blocks.admonition: { }` to your `mkdocs.yaml` file in the section `markdown_extensions`
4. Create a page and add the snippet above (the first one).
5. Reformat with you IDE. I'm using Intelij Idea so it's `CTRL+ALT+L`
6. You should end up with the second snippet. The tabulation disapeared and broke the block attribute

### Version(s) & System Info

Operating System: Non relevant
Python Version: 3.12
Package Version: 
- mkdocs 1.6.1
- mkdocs-material 9.5.34
- mkdocs-material-extensions 1.3.1
- pymdown-extensions 10.11.2

PR #2655, opened 2025-04-30
Feature/alternate block attribute syntax

robinlioret:
Address #2491 

I'm facing the same issue with VSCode auto formatter. It's forcing me to disable formatOnSave, which is not a thing I want.

This PR adds an alternative syntax to the way blocks' attributes header are processed.

It is:
- Completely retrocompatible, so existing project don't have to modify anything.
- Simple yet unlikely to be a content

Per example:
```
/// admonition | Some title
@type: warning

Some content
///
```

### DIFF
--- a/docs/src/markdown/extensions/blocks/index.md
+++ b/docs/src/markdown/extensions/blocks/index.md
@@ -131,6 +131,46 @@ Indented content should always be separated from the block header by one empty l
 option block.
 ///
 
+### Alternative syntax for options
+
+Some formatters tends to remove the leading four spaces before the additional option. In that case, an alternative syntax is possible by replacing the spaces by a single `@` :
+
+/// tab | Alternative syntax
+```text title="Admonition"
+/// admonition | Some title
+@type: warning
+
+Some content
+///
+```
+
+//// html | div.result
+///// admonition | Some title
+@type: warning
+
+Some content
+/////
+////
+///
+
+/// tab | Normal syntax
+```text title="Admonition"
+/// admonition | Some title
+    type: warning
+
+Some content
+///
+```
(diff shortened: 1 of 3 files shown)


facelessuser:
I appreciate the PR, but I'm not sure this is a direction we want to go in.

With Markdown plugins, they will always introduce non-standard syntax, and I don't want to get into the habit of providing alternative syntaxes for every user's personal preferences. I think my comment in the linked issue still stands: https://github.com/facelessuser/pymdown-extensions/issues/2491#issuecomment-2399792024.

robinlioret:
I do agree with you on the will to avoid alternative syntax. It can become messy very quick. I'm actually surprised that so few people encounter this issue.

I think non-standard syntax should be avoided if possible. Extensions should integrate as seamlessly as possible with the language to ensure best user experience. Not sure having the users customizing their Markdown formatter is something we want though. That is something to consider.

However, I won't insist, you've made your point in the issue's comment. 

Please consider formatting standard for the new features. I really like my auto-format on save, very good for consistency across teams.

Should we close this PR without merging ?

robinlioret:
Thanks for the precision. I think I understand the goal. We need a marker for Yaml ish headers that could later be used for other extensions. So it should not interfer with other syntax, but be easy to write and lite to read and parse.

In these conditions @ is not a good solution, I agree.

What about `...`, `:::` or `+>` instead ?

facelessuser:
I'm not looking to use up more unique symbols or symbol combinations that may conflict with some other extension. I'm open to optionally extending the header, similar to what I showed. It's something I had done previously, and we could optionally reinstate it again.

PR #2656, opened 2025-04-30
Add retry mechanism with linear backoff for rate-limited requests in snippets

0xLucca:
This PR adds a retry mechanism to handle rate-limited requests (HTTP 429) when downloading snippets from URLs. The implementation includes:

- New configuration options:
  - `max_retries`: Maximum number of retry attempts (default: 3)
  - `backoff_factor`: Backoff factor for retry attempts (default: 2)

- Linear backoff strategy:
  - Each retry attempt waits for `backoff_factor * attempt` seconds
  - For example, with default settings:
    - First retry: 2 seconds
    - Second retry: 4 seconds
    - Third retry: 6 seconds

- Error handling:
  - Specific handling for HTTP 429 (Too Many Requests) responses
  - Graceful error propagation after all retries are exhausted
  - Maintains existing error handling for other HTTP status codes

This change improves the reliability of snippet downloads when dealing with rate-limited APIs while maintaining backward compatibility with existing configurations.

### DIFF
--- a/docs/src/markdown/extensions/snippets.md
+++ b/docs/src/markdown/extensions/snippets.md
@@ -237,8 +237,7 @@ If either of these is set to zero, the limits will be ignored.
 
 To pass arbitrary HTTP headers in every HTTP request use `url_request_headers`.
 
-/// new | New 10.16
-Rate limiting support was introduced in 10.16. When a URL returns a 429 (Too Many Requests) status code, the extension
+When a URL returns a 429 (Too Many Requests) status code, the extension
 will automatically retry the request with linear backoff. The retry behavior can be configured with:
 
 - `max_retries`: Maximum number of retry attempts (default: 3)
@@ -248,7 +247,6 @@ For example, with default settings, the retry sequence would be:
 - First retry: 2 seconds
 - Second retry: 4 seconds
 - Third retry: 6 seconds
-///
 
 /// warning | Nested Snippets
 One thing to note though, if a snippet is included via a URL, all nested snippets within it must also be URLs. URL
--- a/docs/src/markdown/extensions/snippets.md
+++ b/docs/src/markdown/extensions/snippets.md
@@ -237,6 +237,19 @@ If either of these is set to zero, the limits will be ignored.
 
 To pass arbitrary HTTP headers in every HTTP request use `url_request_headers`.
 
+/// new | New 10.16
+Rate limiting support was introduced in 10.16. When a URL returns a 429 (Too Many Requests) status code, the extension
+will automatically retry the request with linear backoff. The retry behavior can be configured with:
+
+- `max_retries`: Maximum number of retry attempts (default: 3)
+- `backoff_factor`: Backoff factor for retry attempts (default: 2)
+
+For example, with default settings, the retry sequence would be:
+- First retry: 2 seconds
+- Second retry: 4 seconds
+- Third retry: 6 seconds
+///
+
 /// warning | Nested Snippets
 One thing to note though, if a snippet is included via a URL, all nested snippets within it must also be URLs. URL
 snippets are not allowed to reference local snippet files.
@@ -288,3 +301,5 @@ Option                 | Type            | Default          | Description
 `url_request_headers`  | {string:string} | `#!py3 {}`       | Passes arbitrary headers to URL requestor. By default this is set to empty map.
 `dedent_subsections`   | bool            | `#!py3 False`    | Remove any common leading whitespace from every line in text of a subsection that is inserted via "sections" or by "lines".
 `restrict_base_path`   | bool            | `#!py True`      | Ensure that the specified snippets are children of the specified base path(s). This prevents a path relative to the base path, but not explicitly a child of the base path.
+`max_retries`          | int             | `#!py3 3`        | Maximum number of retry attempts when a URL returns a 429 (Too Many Requests) status code.
+`backoff_factor`       | float           | `#!py3 2.0`      | Backoff factor for retry attempts. The wait time between retries is calculated as `backoff_factor * attempt_number` seconds.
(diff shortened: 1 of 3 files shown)


facelessuser:
Interesting. To get merged, we would have to add tests to exercise the code's logic paths (mocking the actual requests). We would need to update docs as well.

I think this could be a good improvement for those who require it.

docs/src/markdown/extensions/snippets.md
```diff
@@ -237,6 +237,19 @@ If either of these is set to zero, the limits will be ignored.
 
 To pass arbitrary HTTP headers in every HTTP request use `url_request_headers`.
 
+/// new | New 10.16
```
facelessuser, line 240: I usually use these "New" blocks just to inform the user when a new feature was added. You should describe the settings outside of the block where we talk about `url_request_headers`, `url_max_size`, and `url_timeout`. Eventually, I'll remove the really old "New" blocks, but the the documentation of the options should remain.

0xLucca:
Hey @facelessuser, I haven’t had a chance to add the tests yet, but I’m aiming to do so this week.

0xLucca:
Hey @facelessuser ! I

I've added the tests and updated the documentation.

For the doc update, I'm not entirely sure if this aligns with what you had in mind. Could you take a look and let me know if any changes are needed?

Regarding the tests, when I run them individually they all pass:
```
python3 -m pytest tests/test_extensions/test_snippets.py::TestURLSnippetsRetry::test_successful_retry
```

However, running the whole class causes some of them to fail:
```
python3 -m pytest tests/test_extensions/test_snippets.py::TestURLSnippetsRetry
```

It seems to be a concurrency related issue or perhaps some shared state between tests

Let me know what you think, and happy to adjust things based on your feedback

facelessuser:
I imagine some of the network stuff is going to have to be mocked. We aren't really trying to unittest the network library, more how we respond to it when it responds with a failure, etc.

I haven't looked too close at it yet, but we don't really need to be doing real network calls, but simulating them so we can make sure we are handling things appropriately based on our expectations of the returns.

0xLucca:
Just to clarify, all the network calls in the tests are already mocked so no real network requests are being made.

pymdownx/snippets.py
```diff
@@ -190,43 +195,65 @@ def download(self, url):
...
+                        content = response.read()
+
+                    # Process lines
+                    last = content.endswith((b'\r', b'\n'))
+                    s_lines = [l.decode(self.encoding) for l in content.splitlines()]
+                    if last:
+                        s_lines.append('')
+                    return s_lines
+
+            except urllib.error.HTTPError as e:
```
facelessuser, line 246: I thought we were checking status earlier, why would we now need to check it here? Are we actually getting a raised exception now instead of just being able to check the status? This isn't inherently bad per se, but I'm curious what's actually getting exercised in the code.

pymdownx/snippets.py
```diff
@@ -190,43 +195,65 @@ def download(self, url):
...
+                if e.code == 429 and attempt < self.max_retries:
+                    wait = self.backoff_factor * attempt
+                    time.sleep(wait)
+                    continue
+                raise SnippetMissingError(f"Cannot download snippet '{url}': {str(e)}")
+            except Exception as e:
```
facelessuser, line 252: This should not be here. Now we are altering the return. I may be raising ValueError as the payload is too big, but now it's being swallowed, retried, and the exception changed. This can cause existing tests to fail. We should break here by just calling `raise` without a new exception. This will just cause the original exception to bubble up.

pymdownx/snippets.py
```diff
@@ -190,43 +195,65 @@ def download(self, url):
...
-            return s_lines
+        for attempt in range(1, self.max_retries + 1):
+            try:
+                http_request = urllib.request.Request(url, headers=self.url_request_headers)
+                timeout = None if self.url_timeout == 0 else self.url_timeout
+                with urllib.request.urlopen(http_request, timeout=timeout) as response:
+                    # Fail if status is not OK
+                    status = response.status if util.PY39 else response.code
+                    
+                    if status == 429:  # Rate limited
```
facelessuser, line 206: This should be simplified as below which should handle all bad status.

```py
if status != 200:
    if status == 429 and attempt < self.max_retries:
        wait = self.backoff_factor * attempt
        time.sleep(wait)
        continue
    raise SnippetMissingError(f"Cannot download snippet '{url}'")
```

Additionally, if we are capturing the rate limited status here, why do we need to check for `urllib.error.HTTPError` later?

facelessuser:
@0xLucca I've added some comments. I suspect because you are altering how exceptions are handled that you are now altering how the extension behaves, causing tests to fail.

pymdownx/snippets.py
facelessuser: I do see in the urlib code that some status can raise an exception. It looks like 429 may do this. If that is the case, then the earlier 429 check is not needed.

pymdownx/snippets.py
facelessuser: I do see in the urlib code that some status can raise an exception. It looks like 429 may do this. If that is the case, then the earlier 429 check is not needed.

facelessuser:
There were a number of issues.

1. How exceptions were being handled.
2. Global config in testing infrastructure was changed, but then the local processor was restored, and the global config was left altered, causing the next test to get unexpected results. Additionally, the order in which tests are run in a Class object is not guaranteed.

facelessuser:
@0xLucca, I refactored your work in PR #2669 fixing various issues. It was just easier for me to quickly resolve since you had it 99% there.

facelessuser:
I went ahead and merged the work. Feel free to create another PR if you find issues or create a bug issue.

PR #2773, opened 2025-10-22
Add inline ID shorthand support for caption/table blocks

yves-chevallier:
## Overview

This PR introduces a shorthand for assigning stable IDs directly in caption headers using either `#my-id` or `id=my-id`.
This allows figures and tables to have persistent identifiers **without relying on YAML front matter**, which can often be confusing due to the required indentation and `ident` boilerplate.

The parser now supports detecting this shorthand **after existing markers** (such as `<`, `>`, or numbering), and validates the identifier using the same logic as `attrs`.
Documentation has been updated accordingly, and regression tests were added to cover both manual and automatic numbering, including `table-caption`.

## Why this helps

This change is **fully backward compatible** while eliminating a common source of indentation errors around `attrs:.`.
It also matches the mental model users already have from *attr-list–style* IDs, making caption syntax cleaner and easier to teach.

In practice, this approach **saves one line** when defining figures or tables with captions.
It’s not a perfect solution, defining a full figure with a caption still takes about five lines — but until a more compact syntax exists, this already makes things a bit better.

The shorthand works consistently across all caption types, so even **custom caption classes** benefit from the same streamlined syntax.

## New syntax example

```md
![Foo](foo.png)

/// figure-caption | #figure-id
Lorem ipsum...
///
```

### DIFF
--- a/tests/test_extensions/test_blocks/test_captions.py
+++ b/tests/test_extensions/test_blocks/test_captions.py
@@ -268,6 +268,44 @@ class TestBlocksCaption(util.MdCase):
             True
         )
 
+    def test_caption_inline_id_with_extra_token(self):
+        """Ensure inline shorthand ID with extra tokens is rejected."""
+
+        self.check_markdown(
+            R"""
+            Paragraph
+            /// figure-caption | #custom-id extra
+            Caption text.
+            ///
+            """,
+            """
+            <p>Paragraph
+            /// figure-caption | #custom-id extra
+            Caption text.
+            ///</p>
+            """,
+            True
+        )
+
+    def test_caption_inline_id_invalid_identifier(self):
+        """Ensure inline shorthand ID with invalid characters is rejected."""
+
+        self.check_markdown(
+            R"""
+            Paragraph
+            /// figure-caption | #invalid!
+            Caption text.
+            ///
+            """,
+            """
+            <p>Paragraph
+            /// figure-caption | #invalid!
+            Caption text.
+            ///</p>
+            """,
+            True
+        )
+
     def test_table_caption_inline_id(self):
         """Test table caption with inline shorthand ID."""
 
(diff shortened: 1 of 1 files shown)


facelessuser:
> The people I work with aren’t very comfortable with editors or complex syntax. I told them Markdown is easy—but in this case, the cumbersome attrs and id syntax, combined with the need to indent attributes exactly (but not the caption), makes it hard for beginners to remember. Markdown is supposed to be simple, which is why I made this proposal.

I'm not saying I won't accept it, but when I'm considering things, I'm also consider how much I want to maintain. So keep that in mind. I do hear what you are saying though.

I haven't looked at the code yet, but I'll say that if we do accept this, I do expect that that standard approach should override the this new id approach if both are used at the same time. Obviously, it would be weird to do both, but I want a predictable approach if we run into this case.

pymdownx/blocks/caption.py
```diff
@@ -236,6 +236,28 @@ def on_validate(self, parent):
...
+                else:
+                    return False
+                if not identifier or tokens:
+                    return False
+                try:
+                    identifier = type_html_identifier(identifier)
+                except Exception:
+                    return False
+                attrs = dict(self.options['attrs'])
+                if 'id' in attrs:
```
facelessuser: This shouldn't fail, the input is good, we just shouldn't override `attrs[id]`. Just return True. A test would need to verify that we do not override the attribute in this case.

pymdownx/blocks/caption.py
```diff
@@ -236,6 +236,28 @@ def on_validate(self, parent):
...
+                    return False
+                attrs['id'] = identifier
+                self.options['attrs'] = attrs
+                argument = ''
+
         if argument:
```
facelessuser: This condition isn't needed anymore as we are handling invalid arguments above.

facelessuser (changes requested):
I think I'm starting to lean towards allowing this, so I've done a formal review. Please review and make the appropriate changes for this to be considered for inclusion.

facelessuser:
Also, if there is missing code coverage within captions after the suggested changes, please create tests to exercise the gaps in test coverage.

tests/test_extensions/test_blocks/test_captions.py
```diff
@@ -226,6 +226,91 @@ def test_manual_prepend(self):
...
+            <figcaption>
+            <p>Table caption text.</p>
+            </figcaption>
+            </figure>
+            ''',
+            True
+        )
+
+    def test_caption_inline_id_attrs_override(self):
+        """Test inline shorthand ID does not override attrs ID."""
```
facelessuser: Surround 'attrs' in backticks so it will be ignored by the spellchecker: `` `attrs` ``.

facelessuser (changes requested):

facelessuser:
@gir-bot lgtm

facelessuser (approved):

PR #2817, opened 2025-12-29
Add custom blockquote extension

### LINKED ISSUE #2814: New Blockquotes extension
### Description

For a long time, it has bothered me how Python Markdown handles blockquotes. For instance, Python Markdown will make one block quote out of the following:

```md
> test

> test
```

This was based on the original markdown spec and was done on purpose for a lazy syntax. Unfortunately, this makes it hard to have two, explicitly separate blockquotes.

We will add a new blockquotes extension that can replace the default one and allow for a more Modern approach used by many other parsers that allows the above example to be treated as two separate blockquotes.

Additionally, since we are adding our own blockquotes extension, we can allow for callouts via GitHub/Obsidian as well. This would take the GitHub or Obsidian syntax and convert them to our familiar admonition and details output format. This would be opt-in.

### Benefits

This would allow for a more sensible input/output format for blockquotes, and for those that like the GitHub/Obsidian style callouts, they could use that as well.

### Solution Idea

_No response_

facelessuser:
Resolves #2814


facelessuser:
@gir-bot lgtm

Issue #2818, opened 2025-12-29
Tasklist: move item text inside label


gromain:
### Description

Hello,

I have a suggestion for improvement of the tasklist extension.
It would be nice to move the text node following a checkbox inside the label html tag. This would make the label clickable in the same way the checkbox is clickable.

<img width="2847" height="1074" alt="Image" src="https://github.com/user-attachments/assets/fb05597a-b1c8-49d4-915f-f91dd8da565a" />

It was suggested before at https://github.com/facelessuser/pymdown-extensions/issues/1109#issuecomment-737754778 , but this issue is completely unrelated to the request back then.

### Benefits

This would make the label clickable!

### Solution Idea

I believe this could be implemented as an additional argument to this method https://github.com/facelessuser/pymdown-extensions/blob/8c7393aafed804c2c67677faf12d720a9dc7dcd8/pymdownx/tasklist.py#L33

gromain:
I use them as interactive checklists. There is no need for further infrastructure for me, as the list is self-contained, once filled the user switch to another task/page. The status of the page doesn't need to be saved and the pages are statically served.

From the tests I ran, images and inline code blocks are not issues (they are clickable too).

However, the behavior could be constrained to text item only so as not to introduce issues.

Would you be open to a pull request implementing this feature (with a specific flag to activate it)?

facelessuser:
I think due to my earlier statement, specifically wanting to target and mimic GitHub functionality, I'm satisfied with the current behavior.

I'm also not personally convinced complex content would work "well". This is subjective, of course. You are free to create a prototype that could be explored, but, because I don't want to waste your time as well, I am already biased to leave it as is, so it may be difficult to convince me otherwise. If knowing this, you are still willing to try and convince me, then feel free 🙂.

PR #2819, opened 2025-12-30
Feature: tasklist with clickable label

gromain:
Hello,

Following our conversation in #2818, I took a stab at this. Let me know what you think about the proposed changes.

I'm not sure how to properly format the text, ruff wanted to do some changes in other places than I've touched (mainly replacing `'` with `"`). Let me know if it's ok to run it and make some changes in places I didn't touch. For now, I didn't use it for clarity of the changes proposed.

For the changelist:
- I first moved the `get_checkbox` function to the class  `TasklistTreeprocessor`, to simplify the call to the function.
- I then included the `md.htmlStash.store` call inside the `get_checkbox` method.
- I've expanded the tests to also cover the multiple lines tasklist test cases to all existing tasklist tests.
- Finally, I've added  the `clickable_label` option and set it so that when in use, the label include whatever markdown code (or html) is included in the task item.

I didn't change the documentation yet, and probably there are a couple of formatting issue, hence the draft on this pull request.

### DIFF
--- a/pymdownx/tasklist.py
+++ b/pymdownx/tasklist.py
@@ -66,16 +66,27 @@ class TasklistTreeprocessor(Treeprocessor):
         state = match.group('state')
 
         if self.custom_checkbox:
-            return self.md.htmlStash.store(
-                '<label class="task-list-control">' +
-                '<input type="checkbox"{}{}/>'.format(
-                    ' disabled' if not self.clickable_checkbox else '',
-                    ' checked' if state.lower() == 'x' else '') +
-                '<span class="task-list-indicator"></span></label> '
-            ) + match.group('line')
-        return self.md.htmlStash.store('<input type="checkbox"{}{}/> '.format(
-            " disabled" if not self.clickable_checkbox else "", " checked" if state.lower() == "x" else ""
-        )) + match.group('line')
+            if self.clickable_label:
+                line = self.md.htmlStash.store(
+                    '<label><span class="task-list-control">' +
+                    '<input type="checkbox"{}{}/>'.format(
+                        ' disabled' if not self.clickable_checkbox else '',
+                        ' checked' if state.lower() == 'x' else '') +
+                    '<span class="task-list-indicator"></span></span>'
+                ) + match.group('line') + self.md.htmlStash.store("</label>")
+            else:
+                line = self.md.htmlStash.store(
+                    '<label class="task-list-control">' +
+                    '<input type="checkbox"{}{}/>'.format(
+                        ' disabled' if not self.clickable_checkbox else '',
+                        ' checked' if state.lower() == 'x' else '') +
+                    '<span class="task-list-indicator"></span></label> '
+                ) + match.group('line')
+        else:
+            line = self.md.htmlStash.store('<input type="checkbox"{}{}/> '.format(
+                " disabled" if not self.clickable_checkbox else "", " checked" if state.lower() == "x" else ""
+            )) + match.group('line')
+        return line
 
     def run(self, root):
         """Find list items that start with [ ] or [x] or [X]."""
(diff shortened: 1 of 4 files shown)


facelessuser:
I will take a look when I get some time. I will state that I appreciate you putting it under a switch, as that at least helps its chances.

I am curious why the stash is needed. If you could elaborate further on that, it would help.

While I won't promise acceptance right now, I will try to approach this with an open mind. It may require changes if the behavior in some cases is deemed to be awkward.

I know right now I have concerns about certain content nested under list times, and I know you mentioned possibly leaving those out. Does this change leave them out? Or did we want to explore how well they work included first?

facelessuser:
For proper HTML, it seems the label element is inline and should not contain block elements. This also greatly limits how it _should_ be used when speaking about placing all the content of a list item under it.

facelessuser:
I'm not sure what's going on with your setup. In the lint environment, there is no need to install `pymdownx-lexers`, what you see failing. That is only needed for documentation. It is also not needed when building the wheel for pymdown-extensions either. We can peek at the tox environment and see that `pymdownx-lexers` is not installed.


Screenshot 2025-12-30 at 2 23 10 PM

I will assume that there is something wonky with your VS code environment. If no issues are seen in CI here, then it is fine.

facelessuser:
On a side note, it appears when lists are loose, that `<label>` is created under the paragraph, so that is good. It seems it will only apply to the first paragraph, which makes sense.

I also noticed when `custom_checkbox` is disabled, things are no longer clickable. This is likely because we only added labels for styling purposes with custom tasklists, and normal tasklists have not been updated to respect `clickable_label`. Or it is assumed because no labels are generated originally, that there is no need to add clickable labels there?

I'm still a bit confused about the stash. I feel like it shouldn't be needed. I'll have to investigate.

facelessuser:
Never mind. I see now, you were simply moving the stash call. Don't worry about it then. If this is accepted (or even if it isn't), I may clean that up.

gromain:
For loose lists, I believe it works as expected, it's already one of the test cases.

I did not try with SuperFences, I'll run it and see how it goes. Will it be enough to add `    pymdownx.superfences:` inside the correct test case `extensions:` list in `tests/extensions/tasklist/tests.yml` ?

gromain:
What's the new approach? I can have a look and try my hand at it! I have known good tests there so I can port that if needed.

If I have to put my hands in tasklist, I might as well upgrade what needs an upgrade!

facelessuser:
I guess thinking about the implementation more. The reason I probably used the stash was that no content was under the label. There was no concern about further post-processing, but now that you are moving content under the label, we should not be putting it in the stash as that will hide the content, and some things may be expecting some post-processing. We would likely need to attach proper etree elements.

pymdownx/tasklist.py
```diff
@@ -75,17 +57,43 @@ def sub_paragraph(self, li):
...
+
+        if self.custom_checkbox:
+            if self.clickable_label:
+                line = self.md.htmlStash.store(
+                    '<label><span class="task-list-control">' +
+                    '<input type="checkbox"{}{}/>'.format(
+                        ' disabled' if not self.clickable_checkbox else '',
+                        ' checked' if state.lower() == 'x' else '') +
+                    '<span class="task-list-indicator"></span></span>'
+                ) + match.group('line') + self.md.htmlStash.store("</label>")
```
gromain: There, I break up the stash, so the matched group for the rest of the line gets properly rendered

gromain:
That is why i break up the stash in two parts for the clickable label part of the code. Line 76, I put a comment there.

facelessuser:
This is one of the cumbersome things about Python Markdown, and why it may be more effort to achieve what you're looking for than it is really worth.

PR #2917, opened 2026-06-15
Add support for headings in details summaries

waylan:
I implemented the heading-only behavior as you requested. It wasn't difficult. I just copied and modified the regex for hash-heading to not check for lines before/after as there is never going to be multiple lines within a detail title.

### DIFF
--- a/pymdownx/blocks/details.py
+++ b/pymdownx/blocks/details.py
@@ -5,6 +5,7 @@ from ..blocks import BlocksExtension
 import re
 
 RE_SEP = re.compile(r'[_-]+')
+RE_HEADING = re.compile(r'^(?P<level>#{1,6})(?P<header>(?:\\.|[^\\])*?)#*$')
 
 
 class Details(Block):
@@ -81,7 +82,12 @@ class Details(Block):
         # Create the summary
         if summary is not None:
             s = etree.SubElement(el, 'summary')
-            s.text = summary
+            m = RE_HEADING.match(summary)
+            if m:
+                h = etree.SubElement(s, f'h{len(m.group('level'))}')
+                h.text = m.group('header').strip()
+            else:
+                s.text = summary
 
         return el
 
--- a/tests/test_extensions/test_blocks/test_details.py
+++ b/tests/test_extensions/test_blocks/test_details.py
@@ -76,6 +76,26 @@ class TestBlocksDetails(util.MdCase):
             True
         )
 
+    def test_heading_in_title(self):
+        """Test details with heading in title."""
+
+        self.check_markdown(
+            R'''
+            /// details | ## A Title
+            Some *content*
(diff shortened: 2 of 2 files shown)


facelessuser:
This is a good working prototype. I think the only other thing I'd probably do differently is either enforce a space after `#` for the line to be considered a header (so as not to conflict with plugins that process GitHub issues `#2`, etc.). Either that, or I'd at least have a toggle to enforce/disable such behavior. Personally, I prefer enforcing the space over the more relaxed approach; if there are those who care strongly enough, a toggle may be preferable.

waylan:
I thought about requiring a space but didn't add that because it is not required for heading elsewhere in the syntax. That seemed inconsistent to me. But then I also was thinking there was no other syntax which could conflict. I hadn't thought of github issues. Personally, I am indifferent to the idea and don't see much need for a toggle.

facelessuser:
Like I said, for me, the space is preferred, but I could go with a toggle if people wanted access to something consistent.

My preference is strong enough that I override the default headers with my own plugin https://facelessuser.github.io/pymdown-extensions/extensions/saneheaders/.

I guess I could look for the existence of our own (pydmownx.saneheaders) plugin in the registered plugin list and change the behavior if it is installed. That locks us to default and specifically our header override, which I'm also fine with.

I'm not specifically asking you to do any of this, only noting that I'll likely want to make the space behavior accessible, one way or another, before I pull this in.

waylan:
So obviously Markdown's default has to stay as-is due to many existing documents (despite any preference on our parts). However, for a third-party extension, anything is game. As there should not be any pre-existing documents in which headings without spaces exist in the titles of details blocks, there is no backward compatibility issue. You can do whatever you want. It works as-is for my needs. I always include a space myself, so would be indifferent if you added the requirement without a way to toggle it.

facelessuser:
Yep, sounds good. I'll figure out what direction I want to head in and revisit.

facelessuser:
I'll be adapting this to require a space after hashes. We'll see if we get any requests to have it without. Thanks for doing the initial work.

waylan:
By the way, I forked this and created a repo (https://github.com/waylan/details_block) which adds this functionality as I needed it immediately. I did not and have no intention of releasing it. I am just using the Repo as the dependency rather than a PyPI package. It has worked great for my needs and it does require a space.

facelessuser:
No worries. I'm just slowly gearing up for a non-bugfix release.

facelessuser:
I ended up caving. It will handle normal headers by default, and if it detects our own `saneheaders` extension, which restricts headers to requiring a space after the hash, only then will it require a space. This will allow users to navigate this naturally without being surprised.

facelessuser:
Closing as a duplicate of #2947.

PR #2946, opened 2026-09-07
Complete rewrite of logic for BetterEm, Tilde, Caret, and Mark

facelessuser:
Requires a release of Python Markdown with the following:

- https://github.com/Python-Markdown/markdown/commit/152a16f.

Details:

- Results should be the closest to CommonMark parsing that that Python
  Markdown is capable of.
- Improve performance and avoid nasty pathological cases of bad
  performance. Any issues in performance upstream in Python Markdown
  cannot be addressed at our level in a reasonable way.
- Add notes that there may be unexpected changes in behavior as this
  corrects many BetterEm specific handling to better align with common
  parsers. This improves complex nesting behaviors and performance, but
  if odd complex cases were used in documentation, they would be there
  some of these fixes may change output.
- If used with Python Markdown without the aforementioned commit,
  performance may be worse for some specific cases. The commit is
  required to enable caching, etc.


facelessuser:
@gir-bot lgtm

